PR #20, opened 2017-06-01
Ellipsizes notification summary when it is too long

### LINKED ISSUE #19: Long notification titles/summaries make the indicator wider without a limit
When notification summaries are too long the size of the indicator popover increases to fit all the text, it should wrap its text to a max width instead.

screenshot

santi698:
Wraps notification summary/title when it is too long to fit in the popover instead of making it larger.

Before:
captura de pantalla de 2017-05-30 11-50-09

After:
captura de pantalla de 2017-06-01 10-38-42

Closes #19.

### DIFF
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -62,6 +62,8 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         grid.margin_end = 6;
 
         var title_label = new Gtk.Label ("<b>" + fix_markup (entry_summary) + "</b>");
+        title_label.wrap_mode = Pango.WrapMode.WORD_CHAR;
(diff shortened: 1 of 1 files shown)


danrabbit:
I don't think we want to ellipsize because this could be important information, but we might want to wrap instead

santi698:
I agree, my main pain is that when notifications get too wide, wrapped body lines become too long and hard to read.

santi698:
Do you think the height of the notifications body should be limited too?

donadigo:
The height is already limitted by the AutomaticScrollBox, the only thing we need here is that these should wrap, as Daniel said, we can't hide this information from the user.

santi698:
Great, I'll update the PR and issue, thanks for the quick replies.

PR #22, opened 2017-06-17
Create COPYING

danrabbit:


danrabbit:
Different license ;) That's GPL, this is LGPL

PR #32, opened 2017-12-22
NotificationEntry.vala: Use Granite.DateTime

danrabbit:
Use relative date time util from Granite instead of own util here

### DIFF
--- a/src/Services/Notification.vala
+++ b/src/Services/Notification.vala
@@ -31,14 +31,14 @@ public class Notifications.Notification : Object {
     public uint32 replaces_id;
     public uint32 id;
     public uint32 pid = 0;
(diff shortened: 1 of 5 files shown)


src/Services/Notification.vala
```diff
@@ -38,7 +38,7 @@ public class Notifications.Notification : Object {
     public AppInfo? app_info = null;
 
     public signal void closed ();
-    public signal bool time_changed (TimeSpan span);
+    public signal bool time_changed (DateTime span);
```
jeremypw: Since there is a possible confusion between GLib.DateTime and Granite.DateTime, I suggest the namespace is included.

PR #34, opened 2018-02-08
Use libbamf to recognize applications correctly

### LINKED ISSUE #11: wingpanel  notification displays wrong program
The notification indicator in wingpanel often displays the wrong program as the one responsible for the notification.
Example:
When getting notification from my printer (out of paper, ...) or even "new updates available" from software centre, it is displayed under pyCharm.
Clicking on the notification opens pyCharm.

Launchpad Details: [#LP1616699](https://bugs.launchpad.net/bugs/1616699) Philippe - 2016-08-25 01:25:22 +0000

### LINKED ISSUE #13: Wrong notification title
On every notification, no matter from what app, the title is: "PyCharm". (See screenshot in the attachments)
And sometimes button "Clear all notifications" overlaps last notification

Launchpad Details: [#LP1682638](https://bugs.launchpad.net/bugs/1682638) Tony - 2017-04-13 20:52:53 +0000

donadigo:
Fixes #11.
Fixes #13.

Instead of trying to be really smart about getting the app id from a single app name string, use libbamf to reliably recognize applications that sent the notification. This completely removes the Utils file as any it's logic is no longer needed.

I understand that using libbamf isn't really great here as it's X dependent, but it would be nice to have this issue fixed for Juno before we start moving to Wayland.


peteruithoven:
Would this fix issues like the following? 
- https://github.com/elementary/gala/issues/61
- https://github.com/elementary/wingpanel-indicator-notifications/issues/11
- https://github.com/elementary/wingpanel-indicator-notifications/issues/13

donadigo:
@peteruithoven yes.

tintou:
Even if this is the right solution for X, I feel like it's a little bit late to introduce X dependencies on our products. I'm not really in favor of this but let anyone review this if it has to be included.

danrabbit:
I think as we discussed the other day, the way forward is to be more restrictive about apps that incorrectly identify themselves to the server. The future we should be working toward is one where users have absolute control over which apps are allowed to send notifications, but this can't happen if developers don't identify their apps correctly and if we're trying to work around this.

donadigo:
The bug is not really caused by applications not sending enough info themselves - but our method of detecting by what application the notification was sent (which involves hacky comparasions and they are often wrong) - this is a better solution than the current one.

danrabbit:
If that's the case, I think we should try to fix that without introducing bamf here. We should be using the ID string sent to the notification server and anything that doesn't send that string should be pushed under the "Other" item so it can be disabled.

donadigo:
Can we make this branch like compile with libbamf conditionally and then if libbamf is not present then put notifications without desktop entry provided to "other'? I want to merge this mainly because the current algorithm for determining what app sent a notification is wrong.

tintou:
libbamf is not tailored for future uses, I'm going to close this

PR #49, opened 2018-09-11
Fix status not updating when the indicator was not opened before

### LINKED ISSUE #48: Notifications are not persisted unless the indicator has been opened at least once
_From @ryonakano on September 6, 2018 4:24_

## Actual Behavior & Steps to reproduce

Power on your Juno Beta1 and use some app that uses notifications, like Music or Mail.

screenshot from 2018-09-06 13-06-31

Notifications appear as bubbles but are not stored in the notification indicator, like this screenshot.

screenshot from 2018-09-06 13-06-46

They are stored in the notification indicator after the notification indicator is opened.

## Expected Behavior

Notifications should appear as bubbles and be stored in the notification indicator even if users don't open the notification indicator.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/63291746-notifications-are-not-persisted-unless-the-indicator-has-been-opened-at-least-once?utm_campaign=plugin&utm_content=tracker%2F60476261&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60476261&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #48.

Moves important initialization from `get_widget` to `get_display_widget` so that the indicator can still persist notifications and change disturb state without opening the indicator.

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -30,6 +30,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
     private Gtk.Box? main_box = null;
     private Gtk.ModelButton clear_all_btn;
     private Gtk.Stack stack;
+    private Wingpanel.Widgets.Switch not_disturb_switch;
 
     private NotificationsList nlist;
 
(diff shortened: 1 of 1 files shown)


donadigo:
@Philip-Scott the reason I put this code specifically in `get_display_widget` because the callbacks of the NotifySettings and NotificationManager use the widgets like `nlist` and `dynamic_icon`. This just ensures that even when those callbacks happen before `get_display_widget` and `get_widget`, the code doesn't produce any fatal warnings.

PR #53, opened 2018-10-19
README.md: Remove dbus-glib dependency

worldofpeace:
ping @tintou

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@
 
 You'll need the following dependencies:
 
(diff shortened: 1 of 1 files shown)


danrabbit:
Hm it looks like dbus-glib is [mentioned in the packaging](https://github.com/elementary/wingpanel-indicator-notifications/blob/deb-packaging/debian/control#L8). If this isn't necessary we should make sure to remove it from there as well

PR #95, opened 2020-01-13
Indicator: fix the state restoration

tintou:
This lead to several warnings as `switch_stack` was being called without dynamic_icon even being a thing


tintou:
The current code requires it before adding children to it to show the right view, It can be addressed in a later PR for sure :)

PR #107, opened 2020-04-03
NotificationsList: Make app_entries public get

danrabbit:


src/Widgets/NotificationsList.vala
```diff
@@ -19,7 +19,8 @@ public class Notifications.NotificationsList : Gtk.ListBox {
     public signal void switch_stack (bool show_list);
     public signal void close_popover ();
 
-    private List<AppEntry> app_entries;
+    public unowned List<AppEntry> app_entries { get; private set; }
```
tintou, line 22:
```suggestion
    public GLib.List<AppEntry> app_entries { get; private set; }
```

tintou:
Ah, looks like I was wrong and vala doesn't like non-unowned GLib.List<>, that's weird

PR #109, opened 2020-04-03
NotificationEntry: Copy line length and truncate rules

danrabbit:
Copies rules from https://github.com/elementary/notifications/blob/master/src/Bubble.vala#L102
* Consistent width: always 33ch wide
* Titles ellipsize, not wrap
* Body can be a maximum of 2 lines
* Forcefully remove extra new lines in body if necessary and truncate each line independently 

Also make timestamps dim label as we do in Mail

### DIFF
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -74,7 +74,7 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
                 string[] lines = body.split ("\n");
                 string stripped_body = lines[0] + "\n";
                 for (int i = 1; i < lines.length; i++) {
-                    stripped_body += lines[i].strip () + "";
+                    stripped_body += lines[i].strip () + " ";
                 }
 
                 body_label.label = stripped_body.strip ();
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -40,33 +40,48 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         hexpand = true;
         get_style_context ().add_class (Gtk.STYLE_CLASS_MENUITEM);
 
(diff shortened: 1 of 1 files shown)


src/Widgets/NotificationEntry.vala
```diff
@@ -40,33 +40,48 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
             body_label.wrap_mode = Pango.WrapMode.WORD_CHAR;
-            grid.attach (body_label, 0, 1, 2, 1);
+            body_label.wrap = true;
+            body_label.xalign = 0;
+
+            if ("\n" in body) {
+                string[] lines = body.split ("\n");
+                string stripped_body = lines[0] + "\n";
+                for (int i = 1; i < lines.length; i++) {
+                    stripped_body += lines[i].strip () + "";
```
jeremypw, line 77: Shouldn't the concatenated lines be separated by space rather than an empty string?

danrabbit: Oops, fixed. Good catch!

src/Widgets/NotificationEntry.vala
```diff
@@ -40,33 +40,48 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+
+            if ("\n" in body) {
+                string[] lines = body.split ("\n");
+                string stripped_body = lines[0] + "\n";
+                for (int i = 1; i < lines.length; i++) {
+                    stripped_body += lines[i].strip () + " ";
+                }
+
+                body_label.label = stripped_body.strip ();
+                body_label.lines = 1;
```
jeremypw, line 81: I am not sure why the number of lines is changed to 1 here when it was set to 2 earlier.

jeremypw: Its OK I get it now.

PR #119, opened 2020-07-07
NotificationEntry: Add border and icon

danrabbit:
Starting to work towards a redesign here. I'd like to show images here and eventually action buttons etc. So it makes sense to have these notifications look more like actual notifications

Screenshot from 2020-07-06 18 19 32@2x

### DIFF
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -37,25 +37,38 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
     }
 
     construct {
-        hexpand = true;
-        get_style_context ().add_class (Gtk.STYLE_CLASS_MENUITEM);
+        var app_icon = notification.app_icon;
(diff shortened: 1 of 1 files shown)


cassidyjames:
I like the direction, but my first thought is that the app headings might be too big. What about dropping the app icon size to 16px (since we get a lot of visual noise with the notification style and the notification's icon), and maybe using something like an `h4`style on the app name instead of `h3`?

danrabbit:
I think h4 might be a little too understated here because the notification title is already bold:

Screenshot from 2020-07-09 12 11 21@2x

But how do you feel about dropping the icon and sticking with h3?

Screenshot from 2020-07-09 12 12 00@2x

cassidyjames:
@danrabbit yeah I can dig H3 and no icon. Especially since most notifications will be with app icons for now, anyway. If we feel the need to bring back app icons in the headings later, we can revisit it.

cassidyjames (approved):

PR #124, opened 2020-09-04
NotificationsList: simplify add_entry_internal

danrabbit:
It turns out we run the same loop twice in these single-use methods. This should make it easier to follow what this method actually does

### DIFF
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -58,14 +58,14 @@ public class Notifications.NotificationsList : Gtk.ListBox {
                 }
             });
 
-            if (!(entry.notification.desktop_id in desktop_id_list)) {
+            if (app_entry == null) {
                 app_entry = new AppEntry (entry);
 
                 app_entries.append (app_entry);
                 prepend (app_entry);
                 insert (entry, 1);
                 table.insert (app_entry.app_info.get_id (), 0);
-            } else if (app_entry != null) {
+            } else {
                 resort_app_entry (app_entry);
                 app_entry.add_notification_entry (entry);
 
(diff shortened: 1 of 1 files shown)


src/Widgets/NotificationsList.vala
```diff
@@ -37,17 +37,43 @@ public class Notifications.NotificationsList : Gtk.ListBox {
...
+                var app_id = _app_entry.app_info.get_id ();
+
+                desktop_id_list += app_id;
+
+                if (app_id == entry.notification.desktop_id && app_entry == null) {
+                    app_entry = _app_entry;
+                }
+            });
+
+            if (!(entry.notification.desktop_id in desktop_id_list)) {
```
jeremypw, line 61: Isn't this equivalent to `app_entry == null`?  We have already scanned the desktop id list.  If it had contained the notification desktop_id then app_entry would have been set.

src/Widgets/NotificationsList.vala
```diff
@@ -37,17 +37,43 @@ public class Notifications.NotificationsList : Gtk.ListBox {
...
+                app_entry = new AppEntry (entry);
+
+                app_entries.append (app_entry);
+                prepend (app_entry);
+                insert (entry, 1);
+                table.insert (app_entry.app_info.get_id (), 0);
+            } else if (app_entry != null) {
```
jeremypw, line 68: Do not need to test `app_entry` again (see previous comment).

src/Widgets/NotificationsList.vala
danrabbit: Yep, you're right. Thanks for helping me make sense of this :sweat_smile:

jeremypw (approved):
No regressions found

PR #128, opened 2020-09-12
Indicator: Connect to built-in listbox signals

danrabbit:
Instead of manually updating, we can use listbox's built-in add and remove signals to make sure that we're always accurate no matter how the list is changed


jeremypw (approved):
Makes sense. Works as expected

PR #130, opened 2020-09-13
Fix button sensitivity setting for clear all notifications

### LINKED ISSUE #129: Clear all notifications action is disabled even when notifications are present
I'm having trouble capturing a screenshot of the notifications indicator. This issue is for Odin daily.

This happens when I have disabled Terminal notification option "Show missed notifications in notification center" but they still accumulate anyway with the "clear all notifications" action greyed out.

vjr:
Fix #129

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -137,8 +137,6 @@ public class Notifications.Indicator : Wingpanel.Indicator {
             });
 
             settings_btn.clicked.connect (show_settings);
-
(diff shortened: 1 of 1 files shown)


jeremypw (changes requested):
You can remove the `update_clear_all_sensitivity ()` in `get_widget ()` now.

Works as expected - thanks!

jeremypw (approved):
Good to go!

PR #131, opened 2020-09-15
Fix settings scheme for remember option

vjr:
Fix https://github.com/elementary/switchboard-plug-notifications/issues/73


vjr:
@jeremypw seems to be fixed now...

jeremypw (approved):
Fixes the issue! Thanks.

PR #134, opened 2020-09-16
NotificationsList: add_entry cleanups

danrabbit:
* Remove counter which is never used for anything
* Instead of returning if AppEntry is null, move that stuff into the conditional where AppEntry is never null
* Make sure we don't iterate over app_entries longer than we need to. We can escape the loop as soon as we find a matching ID


src/Widgets/NotificationsList.vala
```diff
@@ -44,18 +43,15 @@ public class Notifications.NotificationsList : Gtk.ListBox {
...
-            string[] desktop_id_list = {};
-            app_entries.foreach ((_app_entry) => {
-                var app_id = _app_entry.app_info.get_id ();
+            AppEntry? app_entry = null;
 
-                desktop_id_list += app_id;
-
-                if (app_id == entry.notification.desktop_id && app_entry == null) {
+            foreach (unowned AppEntry _app_entry in app_entries) {
+                if (_app_entry.app_info.get_id () == entry.notification.desktop_id) {
```
tintou, line 50: It might be worth storing this value as `unowned string entry_desktop_id = entry.notification.desktop_id` before entering the loop :slightly_smiling_face:

tintou (approved):

Issue #120, opened 2020-07-08
Use libhandy swipes for iOS like notification dismissal/context items


hanaral:
As far as I can tell, the way that notifications are dismissed is not going to change, even with the skeu branch of the indicator. I would like to make the suggestion that libhandy is used so that notifications have a form of interactivity. 

Since trackpads take much longer to precisely click buttons, I think this is the perfect opportunity to give the notifications some weight and also make it easier for laptop/trackpad users to dismiss any previous notifications.

hanaral:
Since the initial work has been merged I'm going to close, feel free to reopen if you want it to keep track

PR #136, opened 2020-09-24
Swipe gesture to dismiss notifications

fs3o5:
This PR adds a swipe gesture to dismiss notifications. #120 
A swipe in the other direction could perhaps execute the default action?
I wasn't sure wether a swipe to the left or right was better


src/Widgets/NotificationEntry.vala
```diff
@@ -115,6 +139,13 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         });
 
         notification.closed.connect (() => clear ());
+
+        carousel.page_changed.connect ((index) => {
+            if (index == 1) {
+                nlist.remove_entry (this);
```
donadigo, line 145: Instead of creating a circular dependency with the NotificationsList, add a "removed" signal to this class and let the list listen to this signal and remove the notification.

src/Widgets/NotificationsList.vala
```diff
@@ -77,6 +77,19 @@ public class Notifications.NotificationsList : Gtk.ListBox {
...
+        AppEntry? app_entry = null;
+
+        unowned string entry_desktop_id = entry.notification.desktop_id;
+        foreach (unowned AppEntry _app_entry in app_entries) {
+            if (_app_entry.app_info.get_id () == entry_desktop_id) {
+                app_entry = _app_entry;
+                continue;
+            }
+        }
+        app_entry.remove_notification_entry.begin (entry);
```
donadigo, line 90: Ideally, we would like to check if the app entry before use is not null here, even though it should not be.

donadigo (changes requested):
Two comments.

src/Widgets/NotificationsList.vala
fs3o5: I don't really see a reason why app entry should be null.

danrabbit (commented):
Hey this is really slick!

I guess we should probably swipe notifications off the screen instead of towards the center. @cassidyjames does that make sense to you?

I guess we should probably also have some kind of additional feedback/indicator of what the completed swipe will do. I think it's pretty common on mobile that swiping a row will reveal a trash can or an X or something to indicate deletion

fs3o5:
The arrow to dismiss multiple notifications points left. Perhaps another icon would be better if the notifications are swiped right to remove them?

danrabbit:
Ah yeah, that icon is primarily used in entries where the arrow is pointing to the content that is being cleared. In this context it might make sense to switch to `edit-clear-all-symbolic`

hanaral:
Wait wtf someone already implemented it? Fantastic!

src/Widgets/NotificationEntry.vala
```diff
@@ -17,12 +17,16 @@
 
 public class Notifications.NotificationEntry : Gtk.ListBoxRow {
     public signal void clear ();
+    public signal void removeNotificationEntry (NotificationEntry entry);
```
donadigo, line 20: We use `remove_notification_entry` naming convention in signals as well. It is also not needed to explicitly declare the entry parameter in the signal itself, Vala will by default pass it to you if there are no arguments (I believe).
```suggestion
    public signal void remove_notification_entry ();
```

src/Widgets/NotificationEntry.vala
```diff
@@ -48,40 +57,53 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
         var grid = new Gtk.Grid () {
+            hexpand = true,
             column_spacing = 6,
-            margin = 6
+            margin = 12,
+            margin_top = 1,
+            margin_bottom = 11
         };
+
+        grid.get_style_context ().add_class (Granite.STYLE_CLASS_CARD);
```
donadigo, line 88: Cache the style context so we don't end up calling the same method twice.
```
unowned Gtk.StyleContext grid_context = grid.get_style_context ();
grid_context.add_class (Granite.STYLE_CLASS_CARD);
grid_context.add_class (Granite.STYLE_CLASS_ROUNDED);
```

src/Widgets/NotificationEntry.vala
```diff
@@ -115,6 +138,13 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         });
 
         notification.closed.connect (() => clear ());
+
+        carousel.page_changed.connect ((index) => {
+            if (index == 1) {
+                removeNotificationEntry (this);
```
donadigo, line 144:
```suggestion
                remove_notification_entry ();
```

src/Widgets/NotificationsList.vala
```diff
@@ -69,6 +69,8 @@ public class Notifications.NotificationsList : Gtk.ListBox {
                 insert (entry, insert_pos + 1);
             }
 
+            entry.removeNotificationEntry.connect (remove_entry);
```
donadigo, line 72:
```suggestion
            entry.remove_notification_entry.connect (remove_entry);
```

donadigo (dismissed):
Some more comments.

src/Widgets/NotificationEntry.vala
```diff
@@ -48,40 +57,54 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+        };
         time_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
 
         var grid = new Gtk.Grid () {
+            hexpand = true,
             column_spacing = 6,
-            margin = 6
+            margin = 12,
+            margin_top = 1,
+            margin_bottom = 11
```
danrabbit, line 85: Looks like you're trying to fix some off-by-one here. Can you leave a comment about what that is?

fs3o5: I made a quick comparison:
left: margin_bottom = 11, margin_top = 1
mid: margin_bottom = 12, margin_top = 0
right: margin_bottom = 12, margin_top = 12
Untitled
In my opinion the second option is the best, because the space over the first notification isn't as big and the notification isn't cropped. 
I'm not sure if there is a better way to fix the cropping.

danrabbit (approved):
So good! Thanks for working on this :)

PR #139, opened 2020-09-29
AppEntry: animate on delete

danrabbit:


src/Widgets/NotificationEntry.vala
```diff
@@ -158,17 +168,27 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
             if (deck.transition_running == false && deck.visible_child != grid) {
+                dismiss ();
                 remove_notification_entry ();
             }
         });
     }
 
+    public void dismiss () {
+        revealer.reveal_child = false;
+        GLib.Timeout.add (revealer.transition_duration, () => {
```
Exalm, line 186: `notify["child-revealed"]`

danrabbit: Well I mean yeah if you wanna do things the easy way duh. I knew that :p

src/Widgets/NotificationEntry.vala
```diff
@@ -158,17 +167,24 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
         deck.notify["transition-running"].connect (() => {
             if (deck.transition_running == false && deck.visible_child != grid) {
-                remove_notification_entry ();
+                clear ();
             }
         });
     }
 
+    public void dismiss () {
+        revealer.reveal_child = false;
```
Exalm, line 182: Do this after connecting. This will fail if animations are disabled.

src/Widgets/NotificationEntry.vala
```diff
@@ -158,17 +167,24 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+        revealer.notify["child-revealed"].connect (() => {
+            destroy ();
```
Exalm, line 184: Generally it's a good idea to check the status here, although I don't think there are any cases here specifically where it can change to revealed status.

src/Widgets/NotificationEntry.vala
```diff
@@ -141,7 +142,15 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         deck.add (delete_right);
         deck.visible_child = grid;
 
-        add (deck);
+        revealer = new Gtk.Revealer () {
+            reveal_child = true,
+            transition_duration = 195,
```
Exalm, line 147: How specific, but not arguing with this :p

danrabbit: I actually have no idea where this number came from. It looks like this should probably be 200 based on https://github.com/elementary/granite/issues/429

src/Widgets/NotificationEntry.vala
```diff
@@ -158,15 +167,24 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
             if (deck.transition_running == false && deck.visible_child != grid) {
-                remove_notification_entry ();
+                clear ();
+            }
+        });
+    }
+
+    public void dismiss () {
+        revealer.notify["child-revealed"].connect (() => {
+            if (revealer.child_revealed == false) {
```
Exalm, line 183: `!revealer.child_revealer`?

Exalm:
Otherwise lgtm, but I haven't run it, only checked the code.

PR #140, opened 2020-09-29
Notification: GObject style and simplify

danrabbit:
* GObject-style construction
* Bump copyright
* Construct and private setters
* Remove unused properties

### DIFF
--- a/src/Services/Notification.vala
+++ b/src/Services/Notification.vala
@@ -55,7 +55,7 @@ public class Notifications.Notification : Object {
 
     public Notification.from_message (DBusMessage message, uint32 id) {
         var body = message.get_body ();
-        var hints = body.get_child_value (Column.HINTS);
+        hints = body.get_child_value (Column.HINTS);
 
         Object (
             actions: body.get_child_value (Column.ACTIONS).dup_strv (),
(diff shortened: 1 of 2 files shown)


danrabbit:
Convert to draft because we messed up somewhere

src/Services/Notification.vala
```diff
@@ -56,67 +53,56 @@ public class Notifications.Notification : Object {
...
+            summary: summary,
+            timestamp: new GLib.DateTime.from_unix_local (unix_time)
+        );
+    }
 
-        if (app_info == null) {
+    construct {
+        if (desktop_id != "") {
+            // Avoid example.desktop.desktop
+            desktop_id.replace (".desktop", "");
```
tintou, line 98: replace it with either a regex that makes sure that we are only replacing the end of the desktop_id or do something like:
```vala
if desktop_id.has_suffix (".desktop") {
    desktop_id = desktop_id.substring (0, - ".desktop".length)
}
```

jeremypw: But we are adding ".desktop" to the end again in the next line??  It looks like the intention was to remove multiple occurrences of ".desktop" and replace with just one at the end.

Could use DESKTOP_ID_EXT rather than ".desktop"

PR #144, opened 2020-09-29
Notification: Add is_transient property

danrabbit:
Instead of `get_is_valid` function. This allows us to make hints private to the dbus message constructor and makes it more clear what we mean by "valid"

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -152,7 +152,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
 
     private void on_notification_received (DBusMessage message, uint32 id) {
         var notification = new Notification.from_message (message, id);
-        if (!notification.get_is_valid () || notification.app_name in EXCEPTIONS) {
+        if (notification.is_transient || notification.app_name in EXCEPTIONS) {
             return;
         }
 
--- a/src/Services/Notification.vala
+++ b/src/Services/Notification.vala
@@ -84,6 +84,9 @@ public class Notifications.Notification : Object {
             desktop_id = FALLBACK_DESKTOP_ID;
             app_info = new DesktopAppInfo (desktop_id);
         }
+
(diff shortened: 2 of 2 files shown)


jeremypw:
Seems to lose a test `app_info != null`.  Now only testing for transience. Not sure if this matters?

danrabbit:
@jeremypw Yeah I have no idea why this was testing for appinfo in the first place, since having appinfo or not doesn't change whether the notification hints say that it's transient. But in any case, it doesn't seem like it's possible to have null appinfo since we set it in both constructors

jeremypw:
Actually, looking at the code `app_info` is a nullable member of Notification.  There is a fallback in the `.from_message` constructor but not in the `.from_data` constructor.  I would suggest `app_info` be made a non-nullable property which is constructed as the fallback then only replaced with another DesktopAppInfo after testing that it is non-null.  The corresponding member in AppEntry should then be made a non-nullable construct property.  Also there are null checks in NotificationEntry and NotificationsList  which are redundant if `notification.app_info` cannot be null and should be removed.

danrabbit:
@jeremypw yeah the app info stuff definitely needs some work in another branch. I'm not really sure why we create a fallback app info. It probably should be allowed to be null, but this needs some more investigation

Regardless I don't see a reason why we wouldn't respect notification transience depending on the presence of appinfo

jeremypw:
It just that this code will crash at:
```
        if (!((DesktopAppInfo)notification.app_info).get_boolean ("X-GNOME-UsesNotifications")) {
            app_id = "gala-other";
        }
```
 if it receives a notification that is not transient and also has no app_info.  I am not sure how likely that is in normal practice but presumably such a notification could theoretically be crafted.  So it would be best to insert a null check here.

jeremypw (approved):
Null app_info now fixed in master - no regressions found.

PR #145, opened 2020-09-30
Notification: Remove data session property

danrabbit:
We set the timestamp by default on construct so this doesn't appear to do anything


jeremypw (changes requested):
The time_changed signal has only one connection and that is made after the signal is emitted because data_session is true.

I think the connection should be made before the emission of "time_changed" is made otherwise the time label will always be "now" for the first minute even for restored notifications.  

I think this is what "data_session" was supposed to do - so it is needed.

danrabbit:
@jeremypw When a notification is constructed, we already set the timestamp in the label to the correct value. This was fixed in https://github.com/elementary/wingpanel-indicator-notifications/pull/133

jeremypw:
Hm, must have been looking at an old version :shrug:  Sorry 'bout that.

jeremypw (approved):

PR #149, opened 2020-10-21
[Desktop] Per-notification Delete Button

lainsce:
Adds the ability to remove one notification at a time while on Desktop. Other form factors untested.

Looks like this:
Screenshot from 2020-10-21 18 08 21

CAVEAT: Place of activation seems to be small (2 pixels :disappointed: ), and the layout of each notification becomes off... If you have any ideas to help fix this, do voice 'em in the comments/reviews.

### DIFF
--- a/data/NotificationEntry.css
+++ b/data/NotificationEntry.css
@@ -18,6 +18,14 @@
     border-bottom-left-radius: 6px;
 }
 
+.delete-affordance.neutral {
+    border-top-left-radius: 6px;
+    border-bottom-left-radius: 6px;
+    border-top-right-radius: 6px;
+    border-bottom-right-radius: 6px;
+    padding: 6px;
+}
+
 .delete-affordance image,
 .delete-affordance label {
     color: mix(@theme_fg_color, @error_color, 0.6);
--- a/data/NotificationEntry.css
+++ b/data/NotificationEntry.css
@@ -24,18 +24,27 @@
     font-weight: 600;
 }
 
-.delete-affordance-button {
-    background-color: mix(@theme_bg_color, @STRAWBERRY_500, 0.3);
-    box-shadow: none;
-    padding: 6px;
-    border-top-left-radius: 6px;
-    border-bottom-left-radius: 6px;
-    border-top-right-radius: 6px;
-    border-bottom-right-radius: 6px;
+.close {
+    padding: 3px;
+    border: none;
+    border-radius: 100%;
+    background-image:
+        linear-gradient(
+            to bottom,
+            @BLACK_300,
+            @BLACK_500
+        );
+    box-shadow:
+        inset 0 0 0 1px alpha (#fff, 0.02),
+        inset 0 1px 0 0 alpha (#fff, 0.07),
+        inset 0 -1px 0 0 alpha (#fff, 0.01),
+        0 0 0 1px alpha (#000, 0.7),
+        0 1px 2px alpha (#000, 0.16),
+        0 2px 3px alpha (#000, 0.23);
+    margin: 3px;
 }
 
-.delete-affordance-button image,
-.delete-affordance-button label {
-    color: mix(@theme_fg_color, @error_color, 0.6);
-    font-weight: 600;
+.close-image {
+    color: #fff;
+    -gtk-icon-shadow: 0 1px 1px alpha (#000, 0.6);
 }
--- a/data/NotificationEntry.css
+++ b/data/NotificationEntry.css
@@ -44,7 +44,7 @@
     margin: 3px;
 }
 
-.close-image {
+.close image {
     color: #fff;
     -gtk-icon-shadow: 0 1px 1px alpha (#000, 0.6);
 }
--- a/data/NotificationEntry.css
+++ b/data/NotificationEntry.css
@@ -18,16 +18,24 @@
     border-bottom-left-radius: 6px;
 }
 
-.delete-affordance.neutral {
+.delete-affordance image,
+.delete-affordance label {
+    color: mix(@theme_fg_color, @error_color, 0.6);
+    font-weight: 600;
+}
+
+.delete-affordance-button {
+    background-color: mix(@theme_bg_color, @STRAWBERRY_500, 0.3);
+    box-shadow: none;
+    padding: 6px;
     border-top-left-radius: 6px;
     border-bottom-left-radius: 6px;
     border-top-right-radius: 6px;
     border-bottom-right-radius: 6px;
-    padding: 6px;
 }
 
-.delete-affordance image,
-.delete-affordance label {
+.delete-affordance-button image,
+.delete-affordance-button label {
     color: mix(@theme_fg_color, @error_color, 0.6);
     font-weight: 600;
 }
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -194,7 +194,7 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         });
 
         eventbox.leave_notify_event.connect ((event) => {
-            delete_revealer.set_reveal_child (false);
+            delete_revealer.reveal_child = false;
             return Gdk.EVENT_STOP;
         });
 
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -76,6 +76,7 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         var grid = new Gtk.Grid () {
             hexpand = true,
             column_spacing = 6,
+            row_homogeneous = true,
             margin = 12,
             // Box shadow is clipped to the margin area
             margin_top = 1,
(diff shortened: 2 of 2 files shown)


lainsce:
Last commit changed some stuff, makes it look better IMO:

Screenshot from 2020-10-22 00 55 03

martinstork:
Wouldn't a simple X be enough? Takes less space and the purpose is clear.

danrabbit:
This design is going to conflict with notification action buttons. It would be better to add the close button on hover in the top left corner as in the notification server. This would keep the design consistent and familiar

lainsce:
Well, last commit made it look like this:

Screenshot from 2020-10-22 16 53 47

I can rework it to look like the top-right close button, but it'll make the layout broken as I already tested it to look like the bubble close button locally, it'd make the popover jump in size dramatically and make it look not as nice.

lainsce:
As for action buttons for notifications, yeah I understand this conflicts with that, but how to tackle this without breaking the layout suddenly like the rounded gray close buttons do? :thinking:

lainsce:
I got it to look nice!

Screenshot from 2020-10-22 18 03 38

lainsce:
Reduced close icon size makes it look better. Open for reviews! :)

Screenshot from 2020-10-22 18 36 41

data/NotificationEntry.css
```diff
@@ -23,3 +23,28 @@
...
+        inset 0 0 0 1px alpha (#fff, 0.02),
+        inset 0 1px 0 0 alpha (#fff, 0.07),
+        inset 0 -1px 0 0 alpha (#fff, 0.01),
+        0 0 0 1px alpha (#000, 0.7),
+        0 1px 2px alpha (#000, 0.16),
+        0 2px 3px alpha (#000, 0.23);
+    margin: 3px;
+}
+
+.close-image {
```
danrabbit, line 47: Probably don't need to add a new class here. You could select with 
```suggestion
.close image {
```

src/Widgets/NotificationEntry.vala
```diff
@@ -78,17 +80,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            margin_top = 14,
             margin_bottom = 11
         };
 
         unowned Gtk.StyleContext grid_context = grid.get_style_context ();
         grid_context.add_class (Granite.STYLE_CLASS_CARD);
         grid_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
 
+        var delete_image = new Gtk.Image.from_icon_name ("window-close-symbolic", Gtk.IconSize.SMALL_TOOLBAR);
+        delete_image.get_style_context ().add_class ("close-image");
```
danrabbit, line 92: As above, adding this close is not necessary
```suggestion
```

src/Widgets/NotificationEntry.vala
```diff
@@ -78,17 +80,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+        delete_image.get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+
+        var delete_box = new Gtk.Button () {
```
danrabbit, line 95: Is there a reason you didn't do `Gtk.Button.from_icon_name`?

src/Widgets/NotificationEntry.vala
```diff
@@ -78,17 +80,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+        delete_box.get_style_context ().add_class ("close");
+        delete_box.get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+
+        delete_box.clicked.connect (() => {
+            clear ();
+        });
+
+        var delete_revealer = new Gtk.Revealer () {
+            reveal_child = false,
+            transition_duration = 200,
```
danrabbit, line 109: There's a Granite constant for this now :) 
```suggestion
            transition_duration = Granite.TRANSITION_DURATION_CLOSE,
```

src/Widgets/NotificationEntry.vala
```diff
@@ -78,17 +80,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            transition_type = Gtk.RevealerTransitionType.SLIDE_UP
```
danrabbit, line 110: This should be crossfade to match the notification server
```suggestion
            transition_type = Gtk.RevealerTransitionType.CROSSFADE
```

src/Widgets/NotificationEntry.vala
```diff
@@ -78,17 +80,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            halign = Gtk.Align.START,
+            valign = Gtk.Align.START,
+            image = delete_image
+        };
+        delete_box.get_style_context ().add_class ("close");
+        delete_box.get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+
+        delete_box.clicked.connect (() => {
+            clear ();
+        });
```
danrabbit, line 105: If we could keep signals at the end after UI construction that would be nice

src/Widgets/NotificationEntry.vala
```diff
@@ -149,7 +175,27 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+        var overlay = new Gtk.Overlay ();
+        overlay.add (revealer);
+        overlay.add_overlay (delete_revealer);
+
+        var eventbox = new Gtk.EventBox ();
+        eventbox.events |= Gdk.EventMask.ENTER_NOTIFY_MASK &
+                           Gdk.EventMask.LEAVE_NOTIFY_MASK;
+
+        eventbox.enter_notify_event.connect ((event) => {
+            delete_revealer.set_reveal_child (true);
```
danrabbit, line 187: elementary code style is to prefer properties :)
```suggestion
            delete_revealer.reveal_child = true;
```

src/Widgets/NotificationEntry.vala
```diff
@@ -149,7 +175,27 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            return false;
```
danrabbit, line 188: This should use Gdk constants
```suggestion
            return Gdk.EVENT_STOP;
```

src/Widgets/NotificationEntry.vala
```diff
@@ -149,7 +175,27 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+        });
+
+        eventbox.leave_notify_event.connect ((event) => {
+            delete_revealer.set_reveal_child (false);
+            return false;
```
danrabbit, line 193: Same as above

src/Widgets/NotificationEntry.vala
```diff
@@ -149,7 +175,27 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
         };
         revealer.add (deck);
 
-        add (revealer);
+        var overlay = new Gtk.Overlay ();
+        overlay.add (revealer);
+        overlay.add_overlay (delete_revealer);
+
+        var eventbox = new Gtk.EventBox ();
```
danrabbit, line 182: This probably needs to be inside the deck. At the moment it appears to be blocking scroll events

danrabbit (changes requested):
At first glance, this appears to do the things but it looks like it's blocking the swipe away gesture. The event box probably needs to be inside the deck

I left some other comments about code style and some questions as well

data/NotificationEntry.css
lainsce: When I did this, exactly as you suggested, something else blocked me from coloring the icon, having it always be the fg color.

src/Widgets/NotificationEntry.vala
lainsce: See above, I had to have an image with its own provider to be able to color the button properly otherwise it would collide with a hardcoded color value somewhere..

src/Widgets/NotificationEntry.vala
lainsce: Nice ;)

src/Widgets/NotificationEntry.vala
lainsce: Alrighty

src/Widgets/NotificationEntry.vala
lainsce: See above for explanation as to why I did it. Something blocks me from coloring the image if I don't do this.

src/Widgets/NotificationEntry.vala
lainsce: Will do.

src/Widgets/NotificationEntry.vala
lainsce: Will change.

src/Widgets/NotificationEntry.vala
lainsce: I'll move it :)

src/Widgets/NotificationEntry.vala
lainsce: Suddenly, it works.

src/Widgets/NotificationEntry.vala
lainsce: Suddenly, it worked now.

data/NotificationEntry.css
lainsce: It worked now. :confused:

lainsce:
All review points addressed.

src/Widgets/NotificationEntry.vala
```diff
@@ -149,10 +174,30 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            clear ();
+        });
+
+        eventbox.enter_notify_event.connect ((event) => {
+            delete_revealer.reveal_child = true;
+            return Gdk.EVENT_STOP;
+        });
+
+        eventbox.leave_notify_event.connect ((event) => {
+            delete_revealer.set_reveal_child (false);
```
danrabbit, line 197: Missed a spot ;)
```suggestion
            delete_revealer.reveal_child = false;
```

src/Widgets/NotificationEntry.vala
```diff
@@ -69,7 +70,8 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
 
 
         var time_label = new Gtk.Label (Granite.DateTime.get_relative_datetime (notification.timestamp)) {
-            margin_end = 6
+            margin_end = 6,
+            halign = Gtk.Align.END
```
danrabbit, line 74: I'm not certain this does anything since the label is pushed to the end by an hexpand already

src/Widgets/NotificationEntry.vala
```diff
@@ -55,7 +55,8 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
 
         var app_image = new Gtk.Image () {
             icon_name = app_icon,
-            pixel_size = 48
+            pixel_size = 48,
+            valign = Gtk.Align.CENTER
```
danrabbit, line 59: Maybe a leftover from a previous design? Doesn't seem necessary to add this

src/Widgets/NotificationEntry.vala
```diff
@@ -78,17 +80,36 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
         };
 
         unowned Gtk.StyleContext grid_context = grid.get_style_context ();
         grid_context.add_class (Granite.STYLE_CLASS_CARD);
         grid_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
 
+        var delete_image = new Gtk.Image.from_icon_name ("window-close-symbolic", Gtk.IconSize.SMALL_TOOLBAR);
+        delete_image.get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+
+        var delete_box = new Gtk.Button () {
```
danrabbit, line 94: Can we call this `delete_button` instead since it's a Gtk.Button subclass and not a Gtk.Box subclass

danrabbit (changes requested):
Hey this looks pretty freakin' good! Just a few more comments and I think it'll be ready to go

Also, don't forget to update the visible child check on 209 and 216 to be overlay now instead of grid

lainsce:
Now it seems all good to go?

danrabbit (approved):
Works as expected and code is solid. Nice work!

PR #150, opened 2020-10-22
NotificationEntry: support replaces

danirabbit:
Fixes #9

PR #151, opened 2020-10-22
Session: Save/Restore replaces_id

danrabbit:
Turns out we were hardcoding this to `0`


danrabbit:
@jeremypw This one is a requisite for replaces in #150

jeremypw (approved):
Makes sense

PR #152, opened 2020-10-22
Notification: Remove unused expire timeout

danrabbit:
This does not appear to ever be used


jeremypw:
Why do we not honor this?

danrabbit:
I guess because it doesn't appear to be supported by Glib.notification and I'm not sure we have any examples of apps wanting to use this. We could probably add support for it later, but right now it's just dead code

jeremypw:
Ah, OK - I did not realise Glib.Notification does not fully implement the interface.

jeremypw (approved):

PR #153, opened 2020-10-26
Move UsesNotifications check to Notification

danrabbit:
Avoid an issue with possibly null appinfo and only assign fallback ID in one place


jeremypw (approved):
Makes sense

PR #154, opened 2020-10-26
Handle NotificationEntry construction in NotificationsList

danrabbit:
In the future when we support replaces, we don't always want to construct a new NotificationEntry every time we get a notification. Having this internal to the `add_entry` method means the notification list will be able to make a determination about whether to construct a new entry or not


jeremypw (approved):
Makes sense - no regressions found

PR #155, opened 2020-10-26
Reduce spacing between bubbles

danrabbit:
#149 significantly increased spacing between bubbles to accommodate the close button, so this tightens it up just a little


cassidyjames (approved):
:+1:

PR #157, opened 2020-10-29
Notification: add stop_timeout

danrabbit:
This does two things:

1. Makes it easier for me to implement "replaces" which I swear is coming eventually
2. Makes it explicit what the purpose of this thing is and make sure we do it whenever we remove a NotificationEntry


jeremypw (approved):

PR #158, opened 2020-10-29
Fix notification icons

danrabbit:
`app_info` is never null, so this doesn't work as intended

This should fix notification icons for apps with no `app_info` that are trying to set an icon manually and actually fall back to `dialog-information` if none is available. Neat!

One problem you'll experience is that your restored notifications that were sent prior to this branch will all be `dialog-information` because we were never properly setting or storing these icons. But this won't happen for any new icons


jeremypw (approved):
Works as expected.

PR #159, opened 2020-10-29
AppEntry: Support null appinfo

danrabbit:
Doesn't make anything else support `null` app info yet, but theoretically you should be able to construct an AppEntry without app_info now

### DIFF
--- a/src/Widgets/AppEntry.vala
+++ b/src/Widgets/AppEntry.vala
@@ -19,6 +19,7 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
     public signal void clear ();
 
     public NotificationEntry entry { get; construct; }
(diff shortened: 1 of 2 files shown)


src/Widgets/AppEntry.vala
```diff
@@ -37,13 +38,19 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
         var notification = entry.notification;
         app_info = notification.app_info;
 
-        var label = new Gtk.Label (app_info.get_name ());
+        string name = _("Other");
```
tintou, line 41: If `app_info` is unlikely to be NULL, then it's maybe better to assign other in an `else` clause (otherwise we are allocating a string to free it right after)

tintou (approved):

PR #162, opened 2020-10-29
NotificationsList: simplify add_entry

danrabbit:
A little easier with "Hide whitespace changes"

* We don't use app_info here, so it only makes sense to check for desktop_id. But desktop_id always gets a fallback in Notification, so this check does nothing
* We don't need to check if app entry is null twice
* We should only connect that signal if we're creating a new app entry, otherwise it's already connected


jeremypw:
This conflicts with #169.  Need to decide which approach to use.

src/Widgets/NotificationsList.vala
```diff
@@ -47,36 +47,32 @@ public class Notifications.NotificationsList : Gtk.ListBox {
...
+
+            resort_app_entry (app_entry);
+            app_entry.add_notification_entry (entry);
+
+            int insert_pos = table.get (app_entry.app_id);
+            insert (entry, insert_pos + 1);
+        } else {
+            var app_entry = new AppEntry (notification.app_info);
+            app_entry.add_notification_entry (entry);
+            app_entry.add_notification_entry (entry);
```
jeremypw, line 63: This line appears to be a duplicate of the previous one??

jeremypw (changes requested):
Apart from one apparently duplicate line, this makes sense and seems to work well with manually sent notifications. I'll test for a while in real world.

The alternative PR does not seem to be being progressed at present so might as well proceed with this.

jeremypw (approved):
LGTM and no regressions noticed

PR #163, opened 2020-10-29
AppEntry: Make remove notification private

danrabbit:


jeremypw (changes requested):
This seems to stop the "Clear all notifications" button working properly.   The notifications list needs a way of removing the app_entries.  Actually I am getting this issue with master as well :open_mouth:  - investigating.

jeremypw:
The culprit is this commit - 770888a3fe8b7b610622a7f73ff67257a760daa7.  `clear_all ()` just clears the hashmap and no longer results in the notification entries being removed.

jeremypw:
After merging the clear all fix this PR causes a regression so that Clear All no longer works properly. Although the notification indicator shows the "No Notifications" icon, the popover still contains a list of notifications which never got destroyed.

danrabbit:
Yeah, this makes more sense now I think since that code was fixed. Woops

PR #164, opened 2020-10-29
Move timeout to NotificationEntry

danrabbit:
This is purely for UI, so it doesn't make sense to have this in the notification object

### DIFF
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -18,10 +18,10 @@
 public class Notifications.NotificationEntry : Gtk.ListBoxRow {
     public signal void clear ();
 
-    public bool active = true;
     public Notification notification { get; construct; }
 
     private Gtk.Revealer revealer;
+    private uint timeout_id;
 
     private static Gtk.CssProvider provider;
     private static Regex entity_regex;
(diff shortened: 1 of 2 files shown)


tintou:
A tiny bit:
Can you initialize: `timeout_id = 0`
and set the `timeout_id = 0` at the end of the TImeout closure.
Also check for `timeout_id == 0` before doing Source.remove.

src/Widgets/NotificationEntry.vala
```diff
@@ -196,10 +196,9 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
             return Gdk.EVENT_STOP;
         });
 
-        notification.time_changed.connect ((timestamp) => {
-            time_label.label = Granite.DateTime.get_relative_datetime (timestamp);
-
-            return active;
+        timeout_id = Timeout.add_seconds_full (Priority.DEFAULT, 60, () => {
+            time_label.label = Granite.DateTime.get_relative_datetime (notification.timestamp);
+            return true;
```
tintou, line 201: Forget about my previous comment, I haven't understood the function, this would make it more obvious:
```suggestion
            return GLib.Source.CONTINUE;
```

jeremypw (approved):
Make sense, no regression

PR #169, opened 2020-11-06
Use a GLib.ListStore to handle the notifications.

tintou:
Changes:
 * Make the Notification object the center of the logic
 * Avoid the use of `static` and `get_default` when it's not necessary
 * A GLib.ListStore allows to ensure that everything is in sync
 * Make AppEntry a header of the Gtk.ListView
 * Make the most things private or declared in its block if not re-used


jeremypw:
I guess this will make some of the other PRs redundant ...

jeremypw:
Compiles and runs OK. Seems like a good idea.  Changes order of notifications so that most recent at bottom - is that intended?

MASTER
Notifications-master

THIS PR
Notifications-bind_model

jeremypw (changes requested):
New notifications get added to the bottom of the list even if an appropriate header is available so you can get duplicate headers.

New Notification - bind_model

jeremypw:
@tintou Any chance of progressing this one?

PR #170, opened 2020-11-07
AppEntry: Simplify the object and base it around the app_info

tintou:


jeremypw (approved):

PR #175, opened 2020-12-07
Add notifications indicator tooltip

### LINKED ISSUE #173: Add tooltip to indicator icon
## Problem

A tooltip could provide some extra information before clicking or just some quicker information that feels faster to access without clicking at all.

## Proposal

Add a tooltip with the status it is indicating, and maybe the middle-click action:

>**No new notifications**
>Middle-click to enable Do Not Disturb

or

>**13 new notifications from 3 apps**
>Middle-click to enable Do Not Disturb

or

>**Do Not Disturb enabled**
>Middle-click to disable

## Prior Art

We added a tooltip to Applications Menu with a longer description and keyboard shortcuts.

pongloongyeat:
Closes #173.

### Demo
Screenshot from 2020-12-08 03-50-22

<strike>As for showing middle-click to enable/disable DnD, I can't parse a mouse key to Granite.markup_accel_tooltip since it takes **keyboard** accelerators.</strike>

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -227,11 +227,11 @@ public class Notifications.Indicator : Wingpanel.Indicator {
                 dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("No new notifications"));
                 break;
             case 1:
-                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("1 new notification from 1 app"));
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("1 new notification"));
                 break;
             default:
                 /* Anything else */
-                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u new notifications from %i apps".printf(number_of_notifications, nlist.app_entries.size)));
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({"MiddleButton"}, _("%u new notifications from %i apps".printf(number_of_notifications, nlist.app_entries.size)));
                 break;
         }
     }
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -231,7 +231,11 @@ public class Notifications.Indicator : Wingpanel.Indicator {
                 break;
             default:
                 /* Anything else */
-                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i apps".printf (number_of_notifications, nlist.app_entries.size)));
+                if (nlist.app_entries.size > 1) {
+                    dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i apps".printf (number_of_notifications, nlist.app_entries.size)));
+                } else {
+                    dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i app".printf (number_of_notifications, nlist.app_entries.size)));
+                }
                 break;
         }
     }
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -231,11 +231,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
                 break;
             default:
                 /* Anything else */
-                if (nlist.app_entries.size > 1) {
-                    dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i apps".printf (number_of_notifications, nlist.app_entries.size)));
-                } else {
-                    dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i app".printf (number_of_notifications, nlist.app_entries.size)));
-                }
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i %s".printf (number_of_notifications, nlist.app_entries.size, ngettext("app", "apps", nlist.app_entries.size))));
                 break;
         }
     }
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -238,8 +238,8 @@ public class Notifications.Indicator : Wingpanel.Indicator {
                 /// TRANSLATORS: A tooltip text for the indicator representing the number of notifications.
                 /// e.g. "2 notifications from 1 app" or "5 notifications from 3 apps"
                 description = _("%s from %s").printf (
-                    ngettext ("%u notification", "%u notifications", number_of_notifications).printf (number_of_notifications),
-                    ngettext ("%i app", "%i apps", number_of_apps).printf (number_of_apps)
+                    _(ngettext ("%u notification", "%u notifications", number_of_notifications)).printf (number_of_notifications),
+                    _(ngettext ("%i app", "%i apps", number_of_apps)).printf (number_of_apps)
                 );
                 break;
         }
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -238,8 +238,8 @@ public class Notifications.Indicator : Wingpanel.Indicator {
                 /// TRANSLATORS: A tooltip text for the indicator representing the number of notifications.
                 /// e.g. "2 notifications from 1 app" or "5 notifications from 3 apps"
                 description = _("%s from %s").printf (
-                    _(ngettext ("%u notification", "%u notifications", number_of_notifications)).printf (number_of_notifications),
-                    _(ngettext ("%i app", "%i apps", number_of_apps)).printf (number_of_apps)
+                    dngettext ("notifications-indicator", "%u notification", "%u notifications", number_of_notifications).printf (number_of_notifications),
+                    dngettext ("notifications-indicator", "%i app", "%i apps", number_of_apps).printf (number_of_apps)
                 );
                 break;
         }
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -214,7 +214,21 @@ public class Notifications.Indicator : Wingpanel.Indicator {
     }
 
     private void update_tooltip () {
-        dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({Middle-click to enable Do Not Disturb}, _("Test"));
+        int number_of_notifications = 0;
+        int number_of_apps = 0;
+
+        switch (number_of_notifications) {
+            case 0:
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("No new notifications"));
+                break;
+            case 1:
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("1 new notification from 1 app"));
+                break;
+            default:
+                /* Anything else */
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%i new notifications from %i app".printf(number_of_notifications, number_of_apps)));
+                break;
+        }
     }
 }
 
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -222,6 +222,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
 
     private void update_tooltip () {
         uint number_of_notifications = Session.get_instance ().get_session_notifications ().length ();
+        int number_of_apps = nlist.app_entries.size;
 
         string description;
         string accel_label;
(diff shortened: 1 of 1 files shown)


src/Indicator.vala
```diff
@@ -210,6 +218,23 @@ public class Notifications.Indicator : Wingpanel.Indicator {
             warning ("Failed to open notifications settings: %s", e.message);
         }
     }
+
+    private void update_tooltip () {
+        uint number_of_notifications = Session.get_instance ().get_session_notifications ().length ();
+
+        switch (number_of_notifications) {
+            case 0:
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("No new notifications"));
```
cassidyjames, line 227:
```suggestion
                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("No notifications"));
```

src/Indicator.vala
```diff
@@ -210,6 +218,23 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+                break;
+            case 1:
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("1 new notification"));
```
cassidyjames, line 230:
```suggestion
                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("1 notification"));
```

src/Indicator.vala
```diff
@@ -210,6 +218,23 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+                break;
+            default:
+                /* Anything else */
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u new notifications from %i apps".printf (number_of_notifications, nlist.app_entries.size)));
```
cassidyjames, line 234:
```suggestion
                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i apps".printf (number_of_notifications, nlist.app_entries.size)));
```

cassidyjames (commented):
I realized "new" is unnecessary here, as we don't distinguish between old and new notifications.

src/Indicator.vala
```diff
@@ -210,6 +218,23 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+        switch (number_of_notifications) {
+            case 0:
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("No notifications"));
+                break;
+            case 1:
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("1 notification"));
+                break;
+            default:
+                /* Anything else */
+                dynamic_icon.tooltip_markup = Granite.markup_accel_tooltip ({}, _("%u notifications from %i apps".printf (number_of_notifications, nlist.app_entries.size)));
```
cassidyjames, line 234: We might need to be more robust with how we create this string; technically you could have 3 notifications from one app, and this would incorrectly pluralize "app".

pongloongyeat: Ahh didn't think about that. I'll work on that.

danrabbit: You'll want to use ngettext here since plurals work differently in some languages than in English

pongloongyeat:
Middle-click tooltip works, but the styling of the tooltip is hardcoded as `accel_label = """<span weight="600" size="smaller" alpha="75%">Middle-click to enable Do Not Disturb</span>"""`. I'm not sure if this is the way to go :/

I've done a simple branch of Granite to accept mouse accelerator labels so the styling is not hardcoded but no clue which would be the better implementation.

src/Indicator.vala
```diff
@@ -171,17 +172,24 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
         set_display_icon_name ();
+        update_tooltip ();
     }
 
     private void update_clear_all_sensitivity () {
         clear_all_btn.sensitive = nlist.app_entries.size > 0;
+
+        /* Since update_clear_all_sensitivity() is called on
+        each popover remove, we can update our tooltip on
+        each popover remove here. */
```
danrabbit, line 183: This should be "notification bubble" instead of "popover" since popover is a specific thing in Gtk

src/Indicator.vala
```diff
@@ -210,6 +219,30 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+
+    private void update_tooltip () {
+        uint number_of_notifications = Session.get_instance ().get_session_notifications ().length ();
+        string accel_label;
+
+        if (notify_settings.get_boolean ("do-not-disturb")) {
+            accel_label = """<span weight="600" size="smaller" alpha="75%">Middle-click to disable Do Not Disturb</span>""";
+        } else {
+            accel_label = """<span weight="600" size="smaller" alpha="75%">Middle-click to enable Do Not Disturb</span>""";
+        }
```
danrabbit, line 231: for localization, and just to be more DRY, the formatting should be split from the string, so you want something like:
```suggestion
        if (notify_settings.get_boolean ("do-not-disturb")) {
            accel_label = _("Middle-click to disable Do Not Disturb");
        } else {
            accel_label = _("Middle-click to enable Do Not Disturb");
        }

        accel_label = """<span weight="600" size="smaller" alpha="75%">%s</span>""".printf (accel_label);
```

danrabbit (changes requested):
Hardcoding the markup here is probably fine. It might make sense to expose that markup string publicly in Granite, but I'm not sure it needs to be a function.

I left a comment just about making localization a bit more foolproof and being more DRY. Ultimately you probably want to just have that markup be the last thing you do and then add the two strings like:

```
string description = "foo notifications from bar apps"
string accel = "Middle-click to turn on Do Not Disturb"

dynamic_icon.tooltip_markup = """%s\n<span weight="600" size="smaller" alpha="75%">%s</span>""".printf (description, accel_label)
```

src/Indicator.vala
pongloongyeat: Got it! Would it make sense to swap out the quotation marks for styling to single quotation marks? i.e. 

```vala
dynamic_icon.tooltip_markup = "%s\n<span weight='600' size='smaller' alpha='75%'>%s</span>".printf (description, accel_label)
```

That way, the newline character doesn't get ignored since the string uses triple quotations `"""`.

pongloongyeat: Actually, it's fine. I'll just update `accel_label` with its styling and join them up later with `"%s\n%s".printf (description, accel_label` instead.

pongloongyeat:
@danrabbit I've applied the suggestions! Thanks for the tip on localisation.

andirsun:
Hey @pongloongyeat you maybe want to change the pango harcoding for the tooltip, check my PR to gala here : https://github.com/elementary/granite/pull/446

Now with the function
```
Granite.markup_contextual_tooltip (string context, string description)
```
You dont need to harcode the markup.

pongloongyeat:
@andirsun I'll change it when the PR gets approved :)

src/Indicator.vala
```diff
@@ -210,6 +211,37 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+        switch (number_of_notifications) {
+            case 0:
+                description = _("No notifications");
+                break;
+            case 1:
+                description = _("1 notification");
+                break;
+            default:
+                /* Anything else */
+                description = _("%u notifications from %i %s".printf (number_of_notifications, number_of_apps, ngettext ("app", "apps", number_of_apps)));
```
danrabbit, line 239: This should also use nggettext for the word "notifications". Maybe @ryonakano has advice on the best way to structure this for translators since the phrase contains two plurals

danrabbit (changes requested):
Hey this looks way better. I just left one inline comment about localization

src/Indicator.vala
pongloongyeat: Hmmm. I'm currently using switch case statements to choose the tooltip text between

- No notifications
- 1 notification
- `$number_of_notifications` notifications from `$number_of_apps` app/apps.

I don't think using nggettext for notifications is needed since we'll only jump into the third case when there is more than one notification whereas nggettext was used for app/apps since "2 notifications from 1 app" is a possibility as pointed out by @cassidyjames.

Please let me know if I'm missing something.

cassidyjames:
A note regarding the markup, we did just merge it as a constant in https://github.com/elementary/granite/pull/447. So you can pass a string to `Granite.TOOLTIP_SECONDARY_TEXT_MARKUP.printf ()`.

src/Indicator.vala
ryonakano: Yeah, I think we also need to use ngettext for languages that have multiple plural forms like Czech. Also, we may want to left a comment for translators, like this:

```suggestion
                /// TRANSLATORS: A tooltip text for the indicator representing the number of notifications.
                /// e.g. "2 notifications from 1 app" or "5 notifications from 3  apps"
                description = _("%s from %s".printf (
                    ngettext ("%u notification", "%u notifications",  number_of_notifications).printf (number_of_notifications),
                    ngettext ("%i app", "%i apps", number_of_apps).printf (number_of_apps)
                ));
```

pongloongyeat: Alright, thanks for pointing it out

pongloongyeat:
I've applied your suggestions. This should be good to go

ryonakano:
2020-12-17 15 26 48 のスクリーンショット

I tested the branch and found a l10n issue: I translated the tooltip in Japanese on my local but it isn't displayed correctly. The ngettext does not seem to be working as expected; "2 apps" and "3 notifications" in the screenshot should be shown as translated.
How can we fix this issue?

pongloongyeat:
I'll look into it. Would you mind telling me the steps you took to translate the string? I'm kinda new to this

ryonakano:
l attach the translation file I edited: [ja.po.zip](https://github.com/elementary/wingpanel-indicator-notifications/files/5707633/ja.po.zip).
Unzip and place it in the `po` dir.

You can reproduce the issue by:

```bash
cd build/
ninja install
killall wingpanel && killall wingpanel
LANGUAGE=ja wingpanel
```

src/Indicator.vala
```diff
@@ -210,6 +211,41 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+                break;
+            case 1:
+                description = _("1 notification");
+                break;
+            default:
+                /// TRANSLATORS: A tooltip text for the indicator representing the number of notifications.
+                /// e.g. "2 notifications from 1 app" or "5 notifications from 3 apps"
+                description = _("%s from %s").printf (
+                    _(ngettext ("%u notification", "%u notifications", number_of_notifications)).printf (number_of_notifications),
+                    _(ngettext ("%i app", "%i apps", number_of_apps)).printf (number_of_apps)
```
ryonakano, line 242:
```suggestion
                    dngettext ("notifications-indicator", "%u notification", "%u notifications", number_of_notifications).printf (number_of_notifications),
                    dngettext ("notifications-indicator", "%i app", "%i apps", number_of_apps).printf (number_of_apps)
```

This should work

pongloongyeat:
Thanks @ryonakano for the suggestion. Translated strings work now.

src/Indicator.vala
```diff
@@ -210,6 +211,41 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+
+        accel_label = Granite.TOOLTIP_SECONDARY_TEXT_MARKUP.printf (accel_label);
+
+        switch (number_of_notifications) {
+            case 0:
+                description = _("No notifications");
+                break;
+            case 1:
+                description = _("1 notification");
+                break;
```
ryonakano, line 236: Just out of curiously, but is there any reason you set this condition? I feel you can merge this condition with the `default` and say: "1 notification from 1 app".

If we merge this condition we have only 2 conditions in this `switch` statement so we may want to use `if` instead.

pongloongyeat: Because 1 notification from 1 app would sound a little weird (in English I guess). If you have only 1 notification, it's only possible that it is from 1 app so putting 1 notification from 1 app would make the "from 1 app" part sound a little redundant (in English) and so I didn't think I would need to show it to the user but yes I do agree it would simplify this into 2 conditions if we combine them.

ryonakano: Ah you're right, it's redundant indeed… simplicity in the code or in the copy :thinking:

danrabbit (approved):
AFAICT this looks and works as expected. Code is looking super clean. Nice work!

PR #177, opened 2020-12-22
Stop spamming the session_file

### LINKED ISSUE #174: Panel crashes when clearing all notifications
## What Happened
As title says.

## Expected Behavior
Shouldn't crash.

## Steps to Reproduce
1. Open notifications popover
2. Click clear all notifications.

I can't successfully reproduce this 100% of the time but I've been noticing this quite a lot lately.

### Logs

I don't really understand the logs. Seems to be a segmentation fault caused by wingpanel-indicator-power I think?
```bash
(captive-login:81772): dconf-DEBUG: 02:21:00.037: sync
(wingpanel:79242): dconf-DEBUG: 02:23:01.862: watch_fast: "/io/elementary/desktop/wingpanel/power/" (establishing: 0, active: 1)
** (wingpanel:79242): DEBUG: 02:23:01.869: Device.vala:174: Connection to UPower device established
** (wingpanel:79242): DEBUG: 02:23:01.872: Indicator.vala:114: Icon changed to "battery-full-charged-symbolic"
** (wingpanel:79242): DEBUG: 02:23:01.879: Device.vala:174: Connection to UPower device established
** (wingpanel:79242): DEBUG: 02:23:01.921: Device.vala:174: Connection to UPower device established
** (wingpanel:79242): DEBUG: 02:23:01.925: Device.vala:174: Connection to UPower device established
** (wingpanel:79242): DEBUG: 02:23:01.927: DeviceManager.vala:140: Device "/org/freedesktop/UPower/devices/battery_BAT1" registered
** (wingpanel:79242): DEBUG: 02:23:01.933: Monitor.vala:42: Initialising process monitor.
(wingpanel:79242): dconf-DEBUG: 02:23:03.695: change_fast
(wingpanel:79242): dconf-DEBUG: 02:23:03.695: change_notify: /io/elementary/desktop/wingpanel/power/show-percentage
(wingpanel:79242): dconf-DEBUG: 02:23:04.911: change_fast
(wingpanel:79242): dconf-DEBUG: 02:23:04.912: change_notify: /io/elementary/desktop/wingpanel/power/show-percentage
Segmentation fault
```

## Platform Information

eOS6 daily. wingpanel-indicator-network on my own branch at the moment but I've noticed crashes prior to this.

jeremypw:
Fixes #174 

Rather than risk regressions by trying to reduce signal generation during clearing, the session file is protected by throttling calls to the write file function.

As the session file is created by the indicator it is assumed that this does not get changed by anything other than the indicator so it seems unnecessary to keep reading it to determine the number of notification/apps - this information is now obtained from the keyfile object which should be in sync with the session file.


donadigo (commented):
While I'm pretty sure this relaxes the load of writing to a file, I would strongly prefer that we move all of this into GSettings in the future. Then we could simply use the built-in delay mode to handle frequent changes.

jeremypw:
> While I'm pretty sure this relaxes the load of writing to a file, I would strongly prefer that we move all of this into GSettings in the future. Then we could simply use the built-in delay mode to handle frequent changes.

Yes, this is just a quick fix - not wanting to change too much.  But if agreed that it is the right way to go I could work on using GSettings in another PR.

danrabbit:
@jeremypw is this ready for review or are you still working on this?

jeremypw:
@danrabbit TBH I had forgotten about it. Marking for review.

src/Services/Session.vala
```diff
@@ -109,26 +123,34 @@ public class Notifications.Session : GLib.Object {
...
+    //This can get called multiple times in quick succesion. Wait until not called for 150 ms before writing file.
+    private uint write_contents_timeout_id = 0;
+    private bool can_write_contents = false;
     private void write_contents () {
-        try {
-            FileUtils.set_contents (session_file.get_path (), key.to_data ());
-        } catch (FileError e) {
-            warning (e.message);
+        if (write_contents_timeout_id > 0) {
+            can_write_contents = false;
```
davidmhewitt, line 135: I feel it would be cleaner to simply remove the existing timeout, set the id to 0 and not return in this if statement, then there is no need for the extra boolean?

jeremypw: I did not want to keep destroying and recreating the timeout every time a signal arrives within the timeout delay (which could be dozens of times if there are a lot of notifications).  Not sure what overhead that represents, however, and code clarity might override efficiency I guess.

jeremypw:
Closing in favour of #187 as that reduces the number of calls to `write_contents ()` to no more than a couple at a time so its probably not worth throttling.

PR #185, opened 2021-03-05
Use Granite.SwitchModelButton and Gtk.Separator

pongloongyeat:
Requires elementary/granite#474

`Wingpanel.Widgets.Switch` and `Wingpanel.Widgets.Separator` are deprecated.


danrabbit (approved):
Nice work!

PR #189, opened 2021-04-22
Indicator: misc cleanup

danrabbit:
Some fixes from #169 

* Use `unowned var` for style contexts
* Drop support for old gala notifications settings and make settings private
* move `visible` to object
* reduce scope of dnd switch


JoseExposito (approved):
Nice cleanup :+1:

PR #190, opened 2021-04-22
Notification: Don't rewrite valid appinfo

danrabbit:
If we're getting valid appinfo, we shouldn't force apps to be listed under "Other"


davidmhewitt (approved):
Looks sane to me

PR #191, opened 2021-04-22
Notification: make from_data the default constructor

danrabbit:
Setting things up to make this a GObject-style class


ryonakano (approved):
LGTM and make sense.

Issue #176, opened 2020-12-18
Screenshot looks bad with dark mode on GitHub


ryonakano:
![Screenshot from 2020-12-18 14-21-21](https://user-images.githubusercontent.com/26003928/102578278-2553ee00-413d-11eb-9242-fd185bb68784.png)

This is because we take screenshots with "grab the current window" in Screenshot. We may want to take screenshots with selected area with white wallpaper instead.

cassidyjames:
I wonder if we can use CSS to switch to a dark style screenshot? :thinking: But yeah we could probably do _something_ better here.

MrPotatoBobx:
Maybe an opeque panel would fix the issue?

![screenshot](https://user-images.githubusercontent.com/63403023/116792234-03de1800-aac8-11eb-95ef-aa563256239c.png)

ryonakano:
Fixed in the latest release.

PR #192, opened 2021-04-30
Change screenshot background to an opaque color 

MrPotatoBobx:
Fixes https://github.com/elementary/wingpanel-indicator-notifications/issues/176

screenshot with opaque color


cassidyjames:
@MrPotatoBobx I would rather use an actual dark screenshot here instead of putting a solid color around it which will look bad on both light and dark styles. 😅

PR #196, opened 2021-05-14
Show app icons from URIs

danrabbit:
Fixes icons for Chrome and possibly other apps


bluesabre (approved):
Looks good! My past notifications look a lot nicer now. :+1:

PR #197, opened 2021-05-18
Actions: Only run CI on PR

danrabbit:
We don't need to run this again on merge. Use checkout v2 while we're here


jeremypw (approved):
Looks sane.

PR #198, opened 2021-05-19
GResource animation icons

danrabbit:
Since these are specific to this codebase, they should be gresourced


jeremypw (approved):
Makes sense, no regressions seen.

PR #199, opened 2021-05-21
NotificationsList: Don't clear all on misclick

danrabbit:
This is strangely convoluted. The only types of things packed into this listbox are either notifications or app headers, so there's no case where `close == false`. It also doesn't make sense to clear all notifications unless the user clicks the clear all button in the header. So really we only need to check if a notification has been activated here


zeebok (approved):
Yeah, this makes a lot more sense. Was this here from an old idea of having some entries be more interactive?

danrabbit:
Yeah I really don't know what the logic was here. I have a draft branch for actions and all this has to go anyways for that

PR #202, opened 2021-06-11
Support image paths and badges

danrabbit:
Nearly all of this code is copied from the Notifications server. Some tiny adaptations made in Notification and Session

Screenshot from 2021-06-11 10 33 18

Brings us to parity with the notifications server

### DIFF
--- a/src/Services/Notification.vala
+++ b/src/Services/Notification.vala
@@ -22,12 +22,14 @@ public class Notifications.Notification : Object {
     public string app_name;
     public string summary;
     public string message_body;
+    public string image_path { get; private set; default = ""; }
     public string app_icon;
     public string sender;
     public string[] actions;
     public uint32 replaces_id;
     public uint32 id;
     public GLib.DateTime timestamp;
+    public GLib.Icon badge_icon { get; construct set; }
 
     public string desktop_id;
     public DesktopAppInfo? app_info = null;
@@ -52,13 +54,14 @@ public class Notifications.Notification : Object {
     private const string FALLBACK_DESKTOP_ID = "gala-other" + DESKTOP_ID_EXT;
 
     public Notification (
-        uint32 _id, string _app_name, string _app_icon, string _summary, string _message_body,
+        uint32 _id, string _app_name, string _app_icon, string _summary, string _message_body, string _image_path,
         string[] _actions, string _desktop_id, int64 _unix_time, uint64 _replaces_id, string _sender
     ) {
         app_name = _app_name;
         app_icon = _app_icon;
         summary = _summary;
         message_body = _message_body;
+        image_path = _image_path;
         replaces_id = (uint32) _replaces_id;
         id = _id;
         sender = _sender;
@@ -99,6 +102,23 @@ public class Notifications.Notification : Object {
         app_icon = get_string (body, Column.APP_ICON);
         if (app_icon == "" && app_info != null) {
             app_icon = app_info.get_icon ().to_string ();
+        }
+
+        // GLib.Notification.set_icon ()
+        image_path = lookup_string (hints, "image-path");
+        if (image_path != null) {
+            // Ensure we're not being sent symbolic badge icons
+            image_path = image_path.replace ("-symbolic", "");
+
+            // GLib.Notification also sends icon names via this hint
+            if (Gtk.IconTheme.get_default ().has_icon (image_path) && image_path != app_icon) {
+                badge_icon = new ThemedIcon (image_path);
+            }
+
+            var is_a_path = image_path.has_prefix ("/") || image_path.has_prefix ("file://");
+            if (badge_icon != null || !is_a_path) {
+                image_path = "";
+            }
         }
 
         if (app_info == null) {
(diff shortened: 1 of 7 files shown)


src/Services/Notification.vala
```diff
@@ -101,6 +104,23 @@ public class Notifications.Notification : Object {
             app_icon = app_info.get_icon ().to_string ();
         }
 
+        // GLib.Notification.set_icon ()
+        image_path = lookup_string (hints, "image-path");
+        if (image_path != null) {
+            // Ensure we're not being sent symbolic badge icons
+            image_path = image_path.replace ("-symbolic", "");
```
cassidyjames, line 111: Am I being paranoid that this isn't explicitly only replaced when it's at the end? If some icon was `poorly-symbolic-named` this would behave unexpectedly.

danrabbit: no because you can also have `-rtl` variants, so it has to be able to remove `-symbolic` from the "middle" of the string to keep that working as well

cassidyjames: Ah that's right. Yay string based matching specs!

src/Widgets/MaskedImage.vala
```diff
@@ -0,0 +1,65 @@
+/*
+* Copyright 2019 elementary, Inc. (https://elementary.io)
```
JoseExposito, line 2:
```suggestion
* Copyright 2021 elementary, Inc. (https://elementary.io)
```

src/Widgets/NotificationEntry.vala
```diff
@@ -61,6 +57,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
             app_image.icon_name = notification.app_icon;
         }
 
+        var image_overlay = new Gtk.Overlay ();
+        image_overlay.valign = Gtk.Align.START;
+
+        if (notification.image_path != null && notification.image_path != "") {
+            try {
+                var scale = get_style_context ().get_scale ();
+                var pixbuf = new Gdk.Pixbuf.from_file_at_size (notification.image_path, 48 * scale, 48 * scale);
```
JoseExposito, line 66: I think it could be a good idea to move those `48`s to a constant and share it with `MaskedImage.ICON_SIZE` or pass the icon size to the `MaskedImage` constructor instead of using a hardcoded value.

```suggestion
                var pixbuf_size = ICON_SIZE * scale;
                var pixbuf = new Gdk.Pixbuf.from_file_at_size (notification.image_path, pixbuf_size, pixbuf_size);
```

src/Widgets/NotificationEntry.vala
```diff
@@ -61,6 +57,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+
+                var masked_image = new Notifications.MaskedImage (pixbuf);
+
+                app_image.pixel_size = 24;
```
JoseExposito, line 70: The same for this `24`

src/Widgets/NotificationEntry.vala
```diff
@@ -61,6 +57,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+                app_image.halign = app_image.valign = Gtk.Align.END;
+
+                image_overlay.add (masked_image);
+                image_overlay.add_overlay (app_image);
+            } catch (Error e) {
+                critical ("Unable to mask image: %s", e.message);
+
+                app_image.pixel_size = 48;
```
JoseExposito, line 78: The constant will need to be used also here.

src/Widgets/NotificationEntry.vala
```diff
@@ -61,6 +57,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+                image_overlay.add (app_image);
+            }
+        } else {
+            app_image.pixel_size = 48;
```
JoseExposito, line 82: And here.

src/Widgets/NotificationEntry.vala
```diff
@@ -61,6 +57,41 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            image_overlay.add (app_image);
+
+            if (notification.badge_icon != null) {
+                var badge_image = new Gtk.Image.from_gicon (notification.badge_icon, Gtk.IconSize.LARGE_TOOLBAR) {
+                    halign = Gtk.Align.END,
+                    valign = Gtk.Align.END,
+                    pixel_size = 24
```
JoseExposito, line 89: The same with the 24.

JoseExposito (changes requested):

src/Widgets/MaskedImage.vala
danrabbit: This file comes from the notification server, so it doesn't make sense to change its copyright since there are no changes in the file

src/Widgets/NotificationEntry.vala
danrabbit: The MaskedImage has to be hardcoded because it matches the size of the overlayed mask, but I've added constants here in NotificationEntry

JoseExposito (approved):
Great job! :tada:

PR #203, opened 2021-06-11
Support Buttons

danirabbit:
Supersedes #199 

Opening as draft because I can't confirm this is actually working


leolost2605:
Any chances this is still going somewhere? It would be really helpful for the "An app is running in the background" notifications from the flatpak background portal to keep the option to allow/forbid without opening a dialog

danirabbit:
@leolost2605 feel free to take over! I think I got stuck somewhere but I'm not sure where

PR #211, opened 2021-07-15
Add support for image-data

bluesabre:
Expands image support to match the notification server.

- Adds support for `image_path` (Spec version 1.1 of `image-path`)
- Adds support for `image-data` (>= 1.2), `image_data` (= 1.1), and `icon_data` (< 1.1)

Since `image-data` is sent as raw image data, this branch writes a temp file with the image and references it. When the notification is dismissed, the temp file is removed.

Some opportunities for improvement:
- Write the temp file to `~/.cache` so it's in a user-private location. Does GLib have something for generating a temp filename within the cache data folder?
- Scale temp files to the max possible dimensions we might use to have minimal disk impact... 48px at 3x ... 144px?

Please let me know if this is something we're interested in and if this is an appropriate direction.

image


danrabbit (approved):
This seems to work as expected. I think we can save optimizations for a point release. Nice work :)

PR #222, opened 2021-09-08
NotificationEntry: Fall back to title and app name when missing body

### LINKED ISSUE #219: Ensure notification always have a title and body
## Problem

If a notification is sent with only a title, it's a bit off balance:

Screenshot from 2021-08-19 12 32 19

## Proposal

If the notification only contains a title, we should use the title as body and the app name as title

## Prior Art

This is what we do in the server

lorencinibrdev:
Should fix #219.


lorencinibrdev:
Attaching screenshot:

image

src/Widgets/NotificationEntry.vala
```diff
@@ -94,8 +94,14 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
                 image_overlay.add_overlay (badge_image);
             }
         }
+        
+        var entry_title = notification.summary;
+
```
cassidyjames, line 99: Whitespace is failing lint

```suggestion

        var entry_title = notification.summary;

```

src/Widgets/NotificationEntry.vala
```diff
@@ -145,34 +151,37 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
-                string[] lines = body.split ("\n");
-                string stripped_body = lines[0] + "\n";
-                for (int i = 1; i < lines.length; i++) {
-                    stripped_body += lines[i].strip () + " ";
-                }
+        
+        if (entry_body == "") {
+            entry_body = notification.summary;
+        }
+
```
cassidyjames, line 158: Whitespace lint here as well

```suggestion

        if (entry_body == "") {
            entry_body = notification.summary;
        }

```

cassidyjames (dismissed):
Couple of whitespace issues

lorencinibrdev:
Thanks again for the review @cassidyjames. I'm sorry for the trouble.

Is there any way to have these checks tests done before doing a PR?

danrabbit (changes requested):
One more small thing! It looks like `app_name` is guaranteed to be set. If we get valid AppInfo in `notification` we should make sure to set `app_name` there

And yes you could install an run vala-lint ahead of time if you like :)

lorencinibrdev:
Hi @danrabbit , thanks for the tip on vala-lint. Worked fine here.

Could you please clarify what you meant about the PR? I checked here, like you said, seems like app_name is indeed always set on Notification class.

danrabbit:
Sorry I meant to say that it *isnt* guaranteed that `app_name` is set.

If we have valid appinfo, we should make sure to set `app_name` based on that appinfo

lorencinibrdev:
Think i got it now. Thanks for the help! Let me know if any change is necessary.

lorencinibrdev:
Adding a image as evidence.

ScreenShot

danrabbit (approved):
Working as expected, thanks for the fix!

PR #224, opened 2021-09-28
NotificationsMonitor: Replace deprecated eavesdrop monitoring

### LINKED ISSUE #210: Fedora: Unread notifications never going into notification dropdown
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
_'scuse me if this is supposed to be in [elementary/wingpanel-indicator-notifications](https://github.com/elementary/wingpanel-indicator-notifications/issues), or even somewhere else for non-elementaryOS issues (I couldn't quite work out who is in charge of packaging elementary stuff on Fedora)._

With Pantheon all set up on Fedora I noticed whenever I get a notification from an app, the notification bell doesn't do the usual thing of animating and showing a dot over it, nor does the notification itself go into the list after expiring (and the list itself always permanently states "No Notifications").

### To Reproduce
1. Install neccesary packages for a basic Pantheon/elementary install on Fedora
2. Send a notification from `notify-send`
3. Observe the notification not going into the notification list after expiring

### Expected behavior
Expired notification should go into notifications list to read later.

### Screenshots or screen recordings
image
Notification from `notify-send` (I've also tried it with a few other apps)

image
It never appears here!

## Platform Information
image
_(There's a color combo to make your eyes bleed)_

Versions I've tried
 - notifications
   - `0.11.20210311.giteec89e6.fc3` (package currently on Fedora, not that far behind the git)
 - wingpanel-indicator-notifications
   - `2.1.4-4.20210219git73f7da2.fc34` (package current on Fedora)
   - `47fec3e8dd28bafdde1040bf1959ecce69ea0fd6` (current git commit)

davidmhewitt:
Fixes #210 

The eavesdropping method of listening to messages on the bus is deprecated according to the DBus docs https://dbus.freedesktop.org/doc/dbus-specification.html

The replacement is setting a private bus connection up as a "Monitor".

I've tested this restores the functionality of the notification indicator on Fedora, where nothing was happening, presumably due to newer versions of DBus or stricter blocking of the older eavesdropping method. I've also tested that this continues to work on Odin.

I've also replaced the `error` calls with more appropriate things. We shouldn't crash the whole of wingpanel just because notifications aren't working properly.

### DIFF
--- a/src/Services/Interfaces.vala
+++ b/src/Services/Interfaces.vala
@@ -17,23 +17,5 @@
 
 [DBus (name = "org.freedesktop.Notifications")]
 public interface Notifications.INotifications : Object {
-    public signal void notification_closed (uint32 id, uint32 reason);
     public signal void action_invoked (uint32 id, string action);
-    public abstract uint32 notify (string app_name,
-                                uint32 replaces_id,
-                                string app_icon,
-                                string summary,
-                                string body,
-                                string[] actions,
-                                HashTable<string, Variant> hints,
-                                int32 expire_timeout) throws Error;
-}
-
-[DBus (name = "org.freedesktop.DBus")]
-public interface Notifications.IDBus : Object {
-    [DBus (name = "NameHasOwner")]
-    public abstract bool name_has_owner (string name) throws Error;
-
-    [DBus (name = "GetConnectionUnixProcessID")]
-    public abstract uint32 get_connection_unix_process_id (string name) throws Error;
 }
--- a/src/Services/NotificationsMonitor.vala
+++ b/src/Services/NotificationsMonitor.vala
@@ -23,9 +23,9 @@
 public class Notifications.NotificationMonitor : Object {
     private const string NOTIFY_IFACE = "org.freedesktop.Notifications";
     private const string NOTIFY_PATH = "/org/freedesktop/Notifications";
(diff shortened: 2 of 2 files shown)


tintou (approved):

decathorpe:
Thank you for working on this!

The current patch doesn't seem to compile on Fedora 35+ though, and it looks like this might be related to vala 0.53+:

```
../src/Services/NotificationsMonitor.vala:52.30-52.71: error: Access to instance member `GLib.BusType.get_address_sync' denied
            string address = BusType.get_address_sync (BusType.SESSION);
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Compilation failed: 1 error(s), 0 warning(s)
```

davidmhewitt:
@decathorpe What's the version of Vala you have in Fedora 35?

It's this commit (introduced in the 0.53 series) that breaks it https://gitlab.gnome.org/GNOME/vala/-/commit/9d5708923f64b3914aa213ca780973be853a03b7, but I can only target stable versions of Vala with compiler conditionals. So I'd have to target 0.54.

decathorpe:
We have vala 0.54.1 in Fedora 35 stable repositories now.

PR #232, opened 2021-11-23
Add full markup support

### LINKED ISSUE #231: ' doesn't show up properly
' shows as &39; instead

michaelocurtis2:
Should fix #231,needs testing first


tintou:
We should probably use the same solution as in https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/messageList.js#L17

tintou:
Thanks for looking at it!

1902276:
> Thanks for looking at it!

I've had another stab, using the GNOME-shell

1902276:
Whatever fix is used, will need to be duplicated at https://github.com/elementary/notifications/blob/master/src/Notification.vala

jeremypw:
@Michael-Curtis0 Could you create a matching PR for `elementary/notifications` and check that they work together as expected?  Thanks.  Marking this as `draft` for now as it cannot work alone.  Please revert to `review required` when ready.

Issue #237, opened 2021-12-18
Large number of notifications slow down whole panel


minhdanh:
### What Happened?

At boot wingpanel would take about nearly 1 miniute to appear (when the dock and desktop has appeared). After that it randomly becomes very slow to respond to click (to application menu, time and date, or power icon), may be 5 minutes after the click.
At this moment when checking htop, I can see it's consuming a high amount of CPU:

![image](https://user-images.githubusercontent.com/4865412/146627328-6a8a8a96-9854-4cd3-a406-8e9d587155d0.png)

(The CPU of wingpanel is actually always 100% when it doesn't respond)

### Steps to Reproduce

It happens at boot time, and randomly becomes slow to respond after that. I'm not sure how to find a consistence way to reproduce.

### Expected Behavior

Wingpanel displays as soon the desktop is shown, and respond quickly to mouse click (and ctrl space)

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

I tried to use this command:

killall io.elementary.wingpanel && io.elementary.wingpanel
```

And could see these lines (new lines keep being added):
```shell
(io.elementary.wingpanel:386011): io.elementary.wingpanel.notifications-CRITICAL **: 10:18:23.729: NotificationEntry.vala:79: Unable to mask image: Failed to open file “/tmp/wingpanel-G0N6D1.png”: No such file or directory

(io.elementary.wingpanel:386011): io.elementary.wingpanel.notifications-CRITICAL **: 10:18:23.739: NotificationEntry.vala:79: Unable to mask image: Failed to open file “/tmp/wingpanel-UYK2D1.png”: No such file or directory

(io.elementary.wingpanel:386011): io.elementary.wingpanel.notifications-CRITICAL **: 10:18:23.749: NotificationEntry.vala:79: Unable to mask image: Failed to open file “/tmp/wingpanel-V0V7D1.png”: No such file or directory

(io.elementary.wingpanel:386011): io.elementary.wingpanel.notifications-CRITICAL **: 10:18:23.760: NotificationEntry.vala:79: Unable to mask image: Failed to open file “/tmp/wingpanel-7T53D1.png”: No such file or directory

(io.elementary.wingpanel:386011): io.elementary.wingpanel.notifications-CRITICAL **: 10:18:23.770: NotificationEntry.vala:79: Unable to mask image: Failed to open file “/tmp/wingpanel-2RQ9D1.png”: No such file or directory

[...]

Issue #123, opened 2020-08-24
Large amount of notifications causes wingpanel to crash


cowboycodur:
a large amount of notifications will cause wingpanel to crash and not load on reboot. 

Is there a way we can clear notifications from where ever they are stored?

danrabbit:
can you clarify what you mean by "a large amount"? I can't seem to reproduce this

Closing as "incomplete" until this can be reproduced

You can manually clear the notification session file stored at `~/.cache/.notifications.session`

cowboycodur:
I was running backintime which will throw a notification for every file error rsync has, including "permission denied". It was easily over 1000, but im not sure of how many it would take to lock it up.

If you wanted to reproduce it, install backintime have it backup a folder with X number of files (I would start small) but change the permissions on the files so that backintime would get permission denied.

danrabbit:
Okay yup, I can confirm for 2,000 notifications that we have a freeze here

danrabbit:
For the curious a quick way to reproduce is `for n in {1..2000}; do notify-send test test; done`

danrabbit:
Seems like https://github.com/elementary/wingpanel-indicator-notifications/pull/142 helps with this considerably

CliffBrown:
@danrabbit This issues seems to still be a problem. I have loads of notifications that I just don't clear out from discord etc, and it has a severe hit on the performance of the wingpanel and plank. If I manage to clear the notifications service returns to normal. Won't it be possible to timebox notifications or clear them on reboot or something?

nielspeen:
Came here to report the same issue. This happens for me with the latest stable. It causes wingpanel to start very slowly at login and freeze completely sometime later.

Removing ~/.cache/.notifications.session does indeed fix it temporarily, but like @CliffBrown, I don't manually clear or otherwise process notifications. 

I don't think timeboxing is a good solution though. It still allows for a burst of notifications within that time period to freeze wingpanel.

canopusStar:
> can you clarify what you mean by "a large amount"? I can't seem to reproduce this
> 
> Closing as "incomplete" until this can be reproduced
> 
> You can manually clear the notification session file stored at `~/.cache/.notifications.session`

With 4132 notifications my wing panel entered a permanent crash loop (OS v6.1) disabled all widgets on the right hand side of the panel. 

FIX:  rm ~/.cache/.notifications.session
I have an urgent open case located here: https://github.com/elementary/wingpanel/issues/436

[...]

PR #241, opened 2022-03-03
Load notifications more efficiently in a separate thread

jeremypw:
Partially addresses #237 
Partially address #123 

This PR greatly improves the load speed of the wingpanel when there are thousands of sessions notifications.  The wingpanel's appearance is no longer blocked while the notification listbox is filled with entries.  The notification session file is no longer needlessly rewritten.  A faster function for counting the number of session notifications is provided. The function for getting a list of session notifications is sped up.

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -49,11 +49,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
             dynamic_icon.active = true;
 
             nlist = new NotificationsList ();
-
-            var previous_session = Session.get_instance ().get_session_notifications ();
-            previous_session.foreach ((notification) => {
-                nlist.add_entry (notification);
-            });
+            get_session_notifications.begin ();
 
             Gtk.IconTheme.get_default ().add_resource_path ("/io/elementary/wingpanel/notifications");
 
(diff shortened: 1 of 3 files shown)


src/Indicator.vala
```diff
@@ -68,28 +64,43 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
 
+    private async void get_session_notifications () throws ThreadError {
+        dynamic_icon.tooltip_markup = _("Updating session notifications.  Please wait");
+        dynamic_icon.get_style_context ().add_class ("disabled");
+
+        SourceFunc callback = get_session_notifications.callback;
+        ThreadFunc<bool> run = () => {
+            var previous_session = Session.get_instance ().get_session_notifications ();
+            previous_session.foreach ((notification) => {
+                nlist.add_entry (notification, false); // Do not need to write back to file
```
tintou, line 83: You're adding a GTK widget from another thread here, this leads to crashes, you need to get the previous session in another thread but adding them needs to be done in the GTK thread

jeremypw: Thanks for the review!  How much can be done in a separate thread? i.e. can widgets be created if they are not mapped? Is it emission of signals that causes the problem? It is filling the list box that causes most of the delay in the wingpanel appearing otherwise.   Perhaps we need a virtualizing list box??

jeremypw: Note that the `nlist` is not realized or inserted into parent at the point it is updated in the thread and the indicator is disabled so it cannot be realized until the thread is finished.  I did not experience any crashes or error messages during testing.

jeremypw: Now adding session notifications in Idle instead of thread.  Somewhat slower but still reasonably fast.

jeremypw:
Using an Idle loop instead of a thread results in a noticeable delay in the wingpanel indicators appearing (ca 1 second on my machine with 1000 notifications) but this is still a lot faster than previously.

tintou:
I don't think that we can really address the issue with GTK3, we need GTK4 for very large lists

jeremypw:
Maybe we should prompt the user to clear some notifications if the number gets unreasonably large? Or automatically dismiss the oldest ones.

jeremypw:
@tintou Do you think this PR is enough of an improvement to merge for now and iterate later?

danrabbit (approved):
Yeah until the Gtk4 port we won't get good performance out of the list, but this at least is nice not to block panel loading :+1:

PR #243, opened 2022-03-06
Remove notification from list if bubble dismissed; give accurate tooltip count

jeremypw:
At present if a notification bubble is actively dismissed it remains in the wingpanel list and has to be dismissed again to get rid of it. This seems unexpected.  This PR removes notifications from the list if they are actively closed by the user or a  "CloseNotification" method call. 

It is open to question whether this should also happen in reverse i.e. urgent bubbles should disappear if cleared from the wingpanel list (at present they do not) but that is left for another PR (after hiding bubbles when the indicator is open is implemented).

There was existing code in the NotificationsMonitor that appeared to be intended to implement this already but it did not detect NotificationClosed signals.

A superfluous function/signal was removed from Notification; the Indicator detects when a notification is closed.

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -65,8 +65,15 @@ public class Notifications.Indicator : Wingpanel.Indicator {
             dynamic_icon_style_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
 
             var monitor = NotificationMonitor.get_instance ();
-            monitor.notification_received.connect (on_notification_received);
-            monitor.notification_closed.connect (on_notification_closed);
+            Idle.add (() => { // Monitor initialises asynchronously so cannot connect immediately
+                if (monitor.initialized) {
+                    monitor.notification_received.connect (on_notification_received);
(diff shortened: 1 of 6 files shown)


jeremypw:
I have fixed the tooltip count here, but that could be split out if necessary.

src/Indicator.vala
```diff
@@ -225,8 +230,8 @@ public class Notifications.Indicator : Wingpanel.Indicator {
     }
 
     private void update_tooltip () {
-        uint number_of_notifications = Session.get_instance ().count_notifications ();
```
leolost2605, line 228: Can the `count_notifications` method be removed then? It doesn't seem to be used by anything else

leolost2605 (dismissed):
Only one small question otherwise it looks good to me and can confirm it works as intended. Regarding splitting the fix i wouldn't deem it necessary but that's just my opinion.

src/Indicator.vala
jeremypw: Done.

lenemter (approved):
Works perfectly :+1:

PR #255, opened 2023-05-14
Constant indicator width

leolost2605:
- Set `width_request` of `NotificationsList` to 360 (standard size of a notification)
- Prevent the indicator from resizing when all notifications are cleared

Before:
Before (trimmed)

After:
After (trimmed)

### DIFF
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -40,7 +40,6 @@ public class Notifications.NotificationsList : Gtk.ListBox {
 
         activate_on_single_click = true;
         selection_mode = Gtk.SelectionMode.NONE;
-        width_request = 360;
         set_placeholder (placeholder);
         show_all ();
 
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -40,6 +40,7 @@ public class Notifications.NotificationsList : Gtk.ListBox {
 
         activate_on_single_click = true;
         selection_mode = Gtk.SelectionMode.NONE;
+        width_request = 360;
         set_placeholder (placeholder);
         show_all ();
 
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -129,7 +129,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
 
             main_box = new Gtk.Grid ();
             main_box.orientation = Gtk.Orientation.VERTICAL;
-            main_box.width_request = 300;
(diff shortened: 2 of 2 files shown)


src/Widgets/NotificationsList.vala
```diff
@@ -40,6 +40,7 @@ public class Notifications.NotificationsList : Gtk.ListBox {
 
         activate_on_single_click = true;
         selection_mode = Gtk.SelectionMode.NONE;
+        width_request = 360;
```
jeremypw, line 43: Where does the value of `360` come from?  I notice that in `Indicator.vala` , `get_widget ()` returns a box with a `width_request` of` 300`.

leolost2605: Ah right it might be better to change it there then.
360 is usually the width of the NotificationEntries that are the ListBoxRows representing Notifications but I can't say for sure that's always this way. Might this pose a problem?

jeremypw: Looks like its the `width_chars` and `max_width_chars` properties of the title label.

jeremypw: After further investigation, it seems that the elementary notifications server produces a bubble with width `332` (hard coded).  So if you want the notification list to match the bubbles, that value should be used in `Indicator.vala`.  Maybe in future that value could be parameterized as a setting so both projects could be easily kept in sync (?)

leolost2605: > After further investigation, it seems that the elementary notifications server produces a bubble with width `332` (hard coded). So if you want the notification list to match the bubbles, that value should be used in `Indicator.vala`. Maybe in future that value could be parameterized as a setting so both projects could be easily kept in sync (?)

The NotificationEntries are actually wider than that (they still look about the same my guess would be that's because of some CSS stuff which i don't really know much about), we probably would have to reduce the `width_char` then to keep a constant width which doesn't seem optimal.
I would still plead for increasing the `width_request` in Indicator.vala to 360 as this works most of the time (also with bigger text sizes), is very easy and not a very big change as the "default" size is hardcoded already anyway (at 300). The only problem I can see here is that the notifications indicator might take up more screen space which IMHO doesn't do to much harm at 60 units. Another problem is that it's still hardcoded so it might not work anymore when in the future the default sizes of NotificationEntries change.

jeremypw (approved):
This is not perfect but is simple and works in most common situations so I am happy to approve.  A more refined solution can be proposed later if required.

PR #258, opened 2023-06-10
Support Buttons 2

leolost2605:
- Support buttons

~Currently doesn't work~ because it seems emitting signals on a foreign DBus Object (is this correct terminology?)  isn't possible (which would make sense). However it was used here before (for the default action) so maybe I missed something?
A possible solution would maybe be to implement a DBus method in elementary/notifications that then emits the `ActionInvoked` signal on `org.freedesktop.Notifications`.

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -163,6 +163,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
     }
 
     private void on_notification_received (DBusMessage message, uint32 id) {
+        print ("notification received");
         var notification = new Notification.from_message (message, id);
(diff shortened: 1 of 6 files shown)


leolost2605:
I've implemented the mentioned solution and can confirm it is working now. However I'm not sure whether there might be an easier way?

For testing I monitored the notifications DBus using dbus-monitor "interface='org.freedesktop.Notifications'" and then sent notifications with buttons via the NotifyBackground method on the background portal implementation :)

leolost2605:
Closing in favor of an alternative implementation

PR #263, opened 2023-06-29
Support buttons with action group

leolost2605:
- Works with elementary/notifications#193


src/Widgets/NotificationEntry.vala
```diff
@@ -231,6 +265,28 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
 
         show_all ();
 
+        button_release_event.connect (() => {
+            unowned var action_group = get_action_group (ACTION_GROUP_PREFIX);
+            if (action_group.has_action (notification.default_action)) {
+                notification.app_info.launch_action (Notification.DEFAULT_ACTION, new GLib.AppLaunchContext ());
```
Marukesu, line 271: this is wrong, notifications actions aren't desktop actions.

src/Widgets/NotificationEntry.vala
```diff
@@ -231,6 +265,28 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+                    critical ("Unable to launch app: %s", e.message);
+                }
+                clear ();
+            } else {
+                return Gdk.EVENT_STOP;
+            }
+
+            activate ();
+
+            return Gdk.EVENT_STOP;
```
Marukesu, line 287: this cannot be reached.

src/Widgets/NotificationEntry.vala
```diff
@@ -231,6 +265,28 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            } else if (notification.actions.length == 0) {
+                try {
+                    notification.app_info.launch (null, null);
+                } catch (Error e) {
+                    critical ("Unable to launch app: %s", e.message);
+                }
+                clear ();
+            } else {
+                return Gdk.EVENT_STOP;
+            }
```
Marukesu, line 283: a cleaner way to deal with this is to use the `Gtk.Actionable` interface of `Gtk.ListBoxRow`. and deal with the non-action case in the `activate()` signal.

src/Widgets/NotificationEntry.vala
```diff
@@ -186,6 +193,33 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+                    activate ();
+                    clear (false);
+                });
+
+                action_area.pack_end (button);
+
+                if (!action_area_packed) {
+                    grid.attach (action_area, 0, 2, 3);
+                    action_area_packed = true;
+                }
```
Marukesu, line 219: you don't need to do this inside the loop, the actions are already validated and it's known that we will pack the action_area.

src/Widgets/NotificationEntry.vala
```diff
@@ -186,6 +193,33 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+            bool action_area_packed = false;
+            foreach (var action_name in notification.actions_with_label.get_keys ()) {
+                var button = new Gtk.Button.with_label (notification.actions_with_label[action_name]) {
+                    action_name = ACTION_PREFIX + action_name
+                };
+
+                button.clicked.connect (() => {
+                    activate ();
+                    clear (false);
+                });
```
Marukesu, line 212: you don't need this, you already set the `action_name` in the button.

src/Widgets/NotificationEntry.vala
```diff
@@ -47,6 +50,8 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
     }
 
     construct {
+        insert_action_group (ACTION_GROUP_PREFIX, NotificationMonitor.get_instance ().notifications_action_group);
```
Marukesu, line 53: instead of inserting the action group in the entry, insert it in the listbox. child widgets can use action groups of parent widgets.

src/Widgets/NotificationEntry.vala
leolost2605: Hmm I took that from [the previous implementation](https://github.com/elementary/wingpanel-indicator-notifications/blame/fa7956c29251eb8db60e8053e41d772942d594cf/src/Services/Notification.vala#L149). Maybe it's relied upon by an elementary app even though it's not in the spec? Or maybe @danirabbit can give background on that?

src/Widgets/NotificationEntry.vala
leolost2605: Hmm I think it can be reached as the two previous `if/else if` don't return but just continue or did I miss something?

Marukesu: Oh right, i missed that both first if/else if don't return.

src/Widgets/NotificationEntry.vala
leolost2605: But we need this to close the popover and remove the notification don't we?

Marukesu: i expect the bubble to be removed when the server emit ActionInvoked. you need to deal with this case anyway, otherwise bubbles that was already activated before the bubble fade-out will be kept in the indicator.

src/Widgets/NotificationEntry.vala
leolost2605: done

src/Widgets/NotificationEntry.vala
leolost2605: done

src/Widgets/NotificationEntry.vala
leolost2605: Hmm we would still need the button_release event handler to clear the notification, close the popover and launch the app info so there wouldn't be much gained there. Or is there a way I missed?

src/Widgets/NotificationEntry.vala
danirabbit: I’m not sure what is meant by “notifications aren’t desktop actions” but this is meant to support https://valadoc.org/gio-2.0/GLib.Notification.set_default_action.html

Marukesu: > I’m not sure what is meant by “notifications aren’t desktop actions”…

a action name in the `actions` parameter of `org.freedesktop.Notifications.Notify` method aren't a "exported" action (a action in the [org.gtk.Actions](https://wiki.gnome.org/Projects/GLib/GApplication/DBusAPI#org.gtk.Actions) interface in the applications's object path), or a [desktop action](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s11.html).

`GLib.DesktopAppInfo.launch_action()` is meant to launch desktop actions only (can be abused to launch exported actions because of how GLib.Application is written, but won't work for non-GLib.Application applications).

>…but this is meant to support https://valadoc.org/gio-2.0/GLib.Notification.set_default_action.html

when GLib.Notification uses the `org.freedesktop.Notifications` interface, it will use `GLib.Action.print_detailed_name()` for buttons, and `"default"` for the default action. so we will never receive a real action that could be used with `GLib.DesktopAppInfo.launch_action()`, only false positives.

src/Widgets/NotificationEntry.vala
Marukesu: like with the buttons case, i expect the NotificationMonitor to deal with the `ActionInvoked` signal that this will emit, and remove the notification there.

correct me if i'm wrong, but `Gtk.ListBox.row_activated()` is emitted even when using a action_name or activate() no? closing the popover can be handled there.

src/Widgets/NotificationEntry.vala
leolost2605: Oh right I totally forgot about that, thanks for the reminder :)

src/Widgets/NotificationEntry.vala
leolost2605: > correct me if i'm wrong, but `Gtk.ListBox.row_activated()` is emitted even when using a action_name or activate() no? closing the popover can be handled there.

No it seems it doesn't but that's actually quite useful here so we can differentiate between activation via default action or without any action.

src/Widgets/NotificationEntry.vala
leolost2605: It only launches the AppInfo now without an action. I think that's something we want or shouldn't we launch it at all? IMHO it's probably expected to bring the app to foreground when the notification is clicked on regardless of default action or no action.

leolost2605:
Sorry @Marukesu I thought I hadn't requested a review yet :)

src/Widgets/AppEntry.vala
```diff
@@ -72,9 +72,9 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
         entry.clear.connect (remove_notification_entry);
     }
 
-    public void remove_notification_entry (NotificationEntry entry) {
+    public void remove_notification_entry (NotificationEntry entry, bool close) {
```
Marukesu, line 75: `remove_notification_entry()` is only called when the `NotificationEntry.clear()` signal is emitted. so there's no need to have the close argument.

src/Indicator.vala
```diff
@@ -195,14 +196,35 @@ public class Notifications.Indicator : Wingpanel.Indicator {
...
+                            warning ("Unable to launch app: %s", e.message);
+                        }
+                    }
+
+                    item.clear (false);
+                    close ();
+
+                    return;
+                }
+            }
```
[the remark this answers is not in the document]
leolost2605: It uses search now but regarding the action activation that's meant to support 
> It only launches the AppInfo now without an action. I think that's something we want or shouldn't we launch it at all? IMHO it's probably expected to bring the app to foreground when the notification is clicked on regardless of default action or no action.

Not sure whether that can be a problem?

src/Widgets/AppEntry.vala
```diff
@@ -85,7 +85,7 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
     public void clear_all_notification_entries () {
         Notification[] to_remove = {};
         app_notifications.@foreach ((entry) => {
-            entry.dismiss ();
+            entry.dismiss (true);
             to_remove += entry.notification;
         });
```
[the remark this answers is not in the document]
leolost2605: I'd like to keep this as is until #265 actually gets merged which I can imagine might need some UX input. Or maybe even in general we shouldn't 100% depend on the server as errors can always happen? Also just another wild thought: maybe we should even consider merging server and indicator if that's possible, i.e. merging the server into the indicator? Could probably reduce some duplication effort

src/Widgets/NotificationEntry.vala
```diff
@@ -227,13 +249,15 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
+        }
+
         add (eventbox);
 
         show_all ();
 
-        delete_button.clicked.connect (() => {
-            clear ();
-        });
+        delete_button.clicked.connect (() => clear ());
```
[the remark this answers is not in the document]
leolost2605: Same here we probably should keep it until #265 gets merged

src/Widgets/AppEntry.vala
leolost2605: But the `clear` signal can be called with either true or false?

src/Indicator.vala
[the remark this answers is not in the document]
Marukesu: I expect it to be handled by the `on_row_activated` callback already, we only got in this method by having a action being successfully activated in the server.

src/Widgets/NotificationEntry.vala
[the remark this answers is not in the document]
Marukesu: this call `clear(true)` that activate the action, that make `NotificationClosed` be emitted, that make the Indicator call `clear(false)`. we can skip the first clear() call here doing this.

src/Indicator.vala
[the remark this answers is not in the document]
leolost2605: It can't as the `row_activated` signal unfortunately doesn't get emitted when the `action_name` on a ListBoxRow is set and the row is clicked

src/Widgets/AppEntry.vala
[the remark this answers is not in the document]
leolost2605: Nvm I just saw that the server emits the signal always no matter whether the notifications exist or not
I'll update it :)

Marukesu:
> I don't think this depend on the Session removal…

Thinking better, i take this back.

The session file store notifications by id, ids aren't meant to be persistent and are reused by the server after a restart, that makes the indicator unreliable to deal with removing/closing/activating a notifications because of colliding ids.

the fact this doesn't cause much problems currently is because the indicator don't report back notification closing, neither support action activation of any type (the current default action code is broken). but once this get merged theses issues will start to be recurrent.

leolost2605:
Notifications are now stored by an internal id consisting of timestamp.server_id.
If we load a notification from the session file the server_id will be 0 and no signals will be emitted. If actions exist the buttons will be insensitive (I thought about not storing actions at all but i think this might be clearer as otherwise we might end up with a situation like "Do you want to allow this?" and you've got nothing to click on. Another solution might be to not store notifications with actions at all 🤷)
Requires elementary/notifications#211

src/Widgets/NotificationsList.vala
```diff
@@ -129,11 +134,23 @@ public class Notifications.NotificationsList : Gtk.ListBox {
...
-            notification_entry.notification.run_default_action ();
-            notification_entry.clear ();
+            var notification_entry = (NotificationEntry) row;
+
+            if (notification_entry.notification.default_action != null) {
+                unowned var action_group = get_action_group (ACTION_GROUP_PREFIX);
+                action_group.activate_action (notification_entry.notification.default_action, null);
+            } else if (notification_entry.notification.actions.length > 0) {
+                return;
+            }
```
Marukesu, line 144: we actually dropped this check in the daemon bubbles, it makes sense to drop it here too.

src/Widgets/NotificationsList.vala
```diff
@@ -129,11 +134,23 @@ public class Notifications.NotificationsList : Gtk.ListBox {
...
+
+            try {
+                notification_entry.notification.app_info.launch (null, null);
 
-            close_popover ();
+                notification_entry.clear ();
+                close_popover ();
+            } catch (Error e) {
+                warning ("Unable to launch app: %s", e.message);
+            }
```
Marukesu, line 153: this should be in a else block and close_popover() should be called after it.

Marukesu:
> If actions exist the buttons will be insensitive (I thought about not storing actions at all but i think this might be clearer as otherwise we might end up with a situation like "Do you want to allow this?" and you've got nothing to click on.

i like the idea of the deactivated buttons, i think you could drop the actions but not the label from the session file, and simply don't try to validate anything when restoring. that way we don't risk ending re-enabling a button because the new notification for the same id has a matching action name.

leolost2605:
> i like the idea of the deactivated buttons, i think you could drop the actions but not the label from the session file, and simply don't try to validate anything when restoring. that way we don't risk ending re-enabling a button because the new notification for the same id has a matching action name.

The server id will be 0 and therefore the action always invalid so AFAIK there is no way we could re-enable it by accident. If you want me to do it anyway because of a bit less space used I'd be happy to update it but it'd require quite a bit more code because without we only have to handle a single way buttons are given and I don't think it's really worth it 🤷

src/Widgets/NotificationsList.vala
leolost2605: But if it's in a else block it can't do

> It only launches the AppInfo now without an action. I think that's something we want or shouldn't we launch it at all? IMHO it's probably expected to bring the app to foreground when the notification is clicked on regardless of default action or no action.

right? Do we still want it?

Marukesu: We want to call `GLib.AppInfo.launch()` only when there's no default action, the current code always call it.

Marukesu (approved):
LGTM! Nice work.

danirabbit:
:tada: amazing love it

PR #264, opened 2023-07-22
Allow minimizing app sections

### LINKED ISSUE #217: Group notification to improve handling
## Problem

I have tons of notifications per day, mostly from conversation in telegram and whatsapp and Its hard to see them with scroll.
Screenshot from 2021-08-02 23-18-40

## Proposal

We could group  notifications by app or in a range of time (hours per example) and exapand them with with a click. I really like this behavior in mac os. 
macos-big-sur-grouped-notifications-100851426-orig

leolost2605:
- Allow to minimize an app section by clicking on the name and/or the triangle icon

Probably fixes #217 

Screenshot with a minimized section:
Shrink app sections


vjr:
I really like this PR 👍🏻 at least the UI/UX/behaviour - I cannot comment on the code for now.

I was trying something similar except using a `Hdy.Deck` instead of `ListBox` so the last/latest notification would be visible at the top always and the others would be stacked under the deck (if I'm using proper UI lingo) and clicking the deck would animate the top item move to the bottom, revealing the next in the deck, and so on...

This PR indicator remembers (which I like again) each app's state of whether it is expanded or not every time you close and open the indicator BUT if I log out and log back in ALL apps are in expanded state.

Can we have it remember the state across sessions, or if we don't want to save the states, then make it unexpanded/collapsed by default upon fresh login session for all apps and the other category?

Finally, the expand/collapse animation (which results in the main `ListBoxRow` increasing/decreasing in height with the scollbar resizing too?) with a bunch of notifications in a couple of apps and an other category seems not so smooth to me, but acceptable. If this can be smoothened further it would be a bonus!

Thank you for this PR!

leolost2605:
Hey, thanks for your feedback!

Remembering the expanded state across sessions is now implemented.

Regarding the not so smooth animations: The last commit fixes some flickering that might be happening. If you still have the feeling it could be improved I'd be very happy about a screen record of how it looks for you and some more information about what could need some improvements/what you have in mind it should look like :)

jeremypw (changes requested):
This mostly works but it can crash wingpanel if you "Clear All Notifications" while a section is collapsed.

Another improvement would be to have the section expand/collapse with a click anywhere on the header (apart from the clear button) like Code or Files sidebars.  At the moment you have to click on the expander or the text.

leolost2605:
@jeremypw  both should be fixed now.

Marukesu:
one issue with this is that it won't work with AppEntry being a header set with `Gtk.ListBoxRow.set_header()`. one of the changed needed to make the NotificationList work with GLib.ListModel is making the AppEntry a real header.

leolost2605:
I did it like this for the smooth animation which I think is not possible using the `header_func`? 
I think we could still use a ListModel, we just don't bind it directly and do some translation from the notification based model to a app entry based one.

Marukesu:
> I did it like this for the smooth animation which I think is not possible using the `header_func`? Also this would keep compatability with GTK4's ListViews which should fix most of the performance issues with 1000+ notifications.

`Gtk.ListView.header_factory` is the replacement of `Gtk.ListBox.set_header_func()`, so compatibility isn't a issue (given that a Gtk4 port is OS 8 area).

About the animation, you could try to re-use the NotificationEntry internal revealer to make them slide up to the header, hidding one entry per iteration. that would be compatible with set_header_func().

> I think we could still use a ListModel, we just don't bind it directly and do some translation from the notification based model to a app entry based one.

think about the ListView compatibility, a multi-level model instead of a flat model would hurt performance, no? since the view won't keep track of the notifications, only of applications, and if a app entry has more notifications than could be seen. they would still be rendered when it won't with the flat model. i believe Gtk.ColumunView (that use a ListView per column) have a performance issue compared to a flat ListView because of that.

leolost2605:
> `Gtk.ListView.header_factory` is the replacement of `Gtk.ListBox.set_header_func()`, so compatibility isn't a issue (given that a Gtk4 port is OS 8 area).

Oh nice I didn't know that was coming

> think about the ListView compatibility, a multi-level model instead of a flat model would hurt performance, no? since the view won't keep track of the notifications, only of applications, and if a app entry has more notifications than could be seen. they would still be rendered when it won't with the flat model. i believe Gtk.ColumunView (that use a ListView per column) have a performance issue compared to a flat ListView because of that.

I don't really know a lot about gtk performance but shouldn't the individual ListView's for the notifications take care of whether to render their content or not? Also wouldn't that have us resort the whole list everytime a notification gets added while otherwise we would only have to prepend and maybe change the position of an app entry? I'm not sure though whether that might be a big performance hit?

Meanwhile I updated it to not use separate ListBoxes :)

src/Widgets/AppEntry.vala
```diff
@@ -74,23 +76,25 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
-            notification.revealer.reveal_child = expanded;
+            notification.revealer.reveal_child = expander.expanded;
         }
     }
 
     public void add_notification_entry (NotificationEntry entry) {
         app_notifications.prepend (entry);
         entry.clear.connect (remove_notification_entry);
+
+        entry.revealer.reveal_child = expander.expanded;
```
Marukesu, line 97:
```suggestion
        expander.bind_property ("expanded", entry.revealer, "reveal-child", SYNC_CREATE);
```

`GLib.Object` binds are different of `GLib.Settings` ones, they are one sided unless `GLib.BindingFlags.BIDIRECTIONAL` is specified.

that will let you drop the update_expanded_state() callback.

leolost2605: Updated

src/Widgets/AppEntry.vala
```diff
@@ -53,15 +59,23 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
+        expander = new Gtk.Expander (null) {
+            label_widget = new ExpanderLable (name)
         };
-        grid.add (label);
-        grid.add (clear_btn_entry);
 
-        add (grid);
+        var box = new Gtk.Box (HORIZONTAL, 6);
+        box.add (expander);
+        box.add (clear_btn_entry);
```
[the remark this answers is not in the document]
leolost2605: I added it but IMHO it looks better without 🤷

src/Widgets/AppEntry.vala
```diff
@@ -53,15 +59,23 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
         };
         clear_btn_entry.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
 
-        var grid = new Gtk.Grid () {
-            column_spacing = 6
+        expander = new Gtk.Expander (null) {
+            label_widget = new ExpanderLable (name)
```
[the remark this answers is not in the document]
leolost2605: Thank you, CSS certainly isn't my forte :)

data/io.elementary.wingpanel.notifications.gschema.xml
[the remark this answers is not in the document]
leolost2605: I updated it though I'm not sure whether I did it like you meant? Also I wonder what's the advantage there?

src/Widgets/AppEntry.vala
```diff
@@ -42,26 +52,55 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
         };
         label.get_style_context ().add_class (Granite.STYLE_CLASS_H3_LABEL);
 
+        var expander_content = new Gtk.Box (HORIZONTAL, 3);
+        expander_content.add (image);
+        expander_content.add (label);
+
+        expander = new Gtk.ToggleButton () {
+            child = expander_content,
+            can_focus = false
```
Marukesu, line 70: I believe the more recent pattern is to let buttons have focus, for visual confirmation when navigating with the keyboard. but i would like a confirmation from someone from the UX team.

data/io.elementary.wingpanel.notifications.gschema.xml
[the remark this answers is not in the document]
Marukesu: one advantage is storage, the older way would need to have a path/key pair for every application, now we have only one path and key. another is that it's easier to remove state after a entry got cleared.

src/Widgets/AppEntry.vala
leolost2605: @elementary/ux
Btw it seems I can't request a review from ux in this repo?

src/Widgets/AppEntry.vala
```diff
@@ -22,15 +22,25 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
+    private static HashTable<string, bool> headers;
+
+    private Gtk.ToggleButton expander;
+
+    static construct {
+        provider = new Gtk.CssProvider ();
+        provider.load_from_resource ("/io/elementary/wingpanel/notifications/AppEntryExpander.css");
+
+        settings = new Settings ("io.elementary.wingpanel.notifications");
+        headers = (HashTable<string, bool>)settings.get_value ("headers");
```
Marukesu, line 36:
```suggestion
        headers = (HashTable<string, bool>) settings.get_value ("headers");
```

src/Widgets/AppEntry.vala
```diff
@@ -42,26 +52,55 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
+        margin_end = 12;
+        margin_bottom = 3;
+        margin_top = 6;
+        can_focus = false;
+        child = box;
         show_all ();
 
+        if (app_info.get_id () in headers) {
+            expander.active = headers[app_info.get_id ()];
+        }
```
Marukesu, line 97:
```suggestion
        if (app_id in headers) {
            expander.active = headers[app_id];
        }
```

src/Widgets/AppEntry.vala
```diff
@@ -42,26 +52,55 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
+
+        expander.toggled.connect (() => {
+            headers[app_info.get_id ()] = expander.active;
```
Marukesu, line 100:
```suggestion
            headers[app_id] = expander.active;
```

Marukesu (commented):
Some minor nits.

I believe there's a question of if we should keep the state even after there's no more notifications from the application, or remove it.

My expectation in this case would be that the state get removed. since the header is removed from the indicator too.

vjr:
@leolost2605 I just tried this PR again and I'm happy with it from a user perspective - can't comment on any remaining code or UI/UX review suggestions.

One (final?) request from me, if possible, is can the scrolling be made smooth/kinetic using any readily available UI widgets like from Handy or Gtk4? I've only tried with a mouse on my PC but would be cool if at least with a laptop trackpad this can be achieved?

Cheers!

leolost2605:
> One (final?) request from me, if possible, is can the scrolling be made smooth/kinetic using any readily available UI widgets like from Handy or Gtk4? I've only tried with a mouse on my PC but would be cool if at least with a laptop trackpad this can be achieved?

@vjr AFAIK kinetic scrolling should be enabled by default for touch input. If you want you can test it by enabling "Simulate touchscreen" in the gtk inspector to see whether it's what you had in mind.

> I believe there's a question of if we should keep the state even after there's no more notifications from the application, or remove it.
> My expectation in this case would be that the state get removed. since the header is removed from the indicator too.

@Marukesu I've got no strong opinion on that, I changed it. Now it's collapsed by default or should it rather be expanded?

src/Widgets/AppEntry.vala
danirabbit: Yeah it's generally not a good idea not to allow a widget to have focus since that makes it completely inaccessible via keyboard

danirabbit:
> I was trying something similar except using a `Hdy.Deck` instead of `ListBox` so the last/latest notification would be visible at the top always and the others would be stacked under the deck (if I'm using proper UI lingo) and clicking the deck would animate the top item move to the bottom, revealing the next in the deck, and so on...

This is how iOS handles collapsing as well and I really like that. This way you can still see the latest notification from a given app without it flooding the list. It means you can just always keep a noisy app collapsed and never really have to expand the stack if you don't want to

src/Widgets/AppEntry.vala
```diff
@@ -42,26 +52,55 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
         clear_btn_entry.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
 
-        var grid = new Gtk.Grid () {
-            column_spacing = 6
-        };
-        grid.add (label);
-        grid.add (clear_btn_entry);
+        var box = new Gtk.Box (HORIZONTAL, 6);
+        box.add (expander);
+        box.add (new Gtk.Separator (VERTICAL));
```
danirabbit, line 84: Instead of adding a separator here, we should switch the position of the label and the expander arrow:

Screenshot from 2023-08-29 13 12 12

leolost2605:
> This is how iOS handles collapsing as well and I really like that. This way you can still see the latest notification from a given app without it flooding the list. It means you can just always keep a noisy app collapsed and never really have to expand the stack if you don't want to

Sure I might need some help with CSS stuff though if we want to show like a stack with the other notifications maybe like barely visible under the topmost. Also just to be sure I understand it correctly: we keep the header with the arrow so when clicking it, it will still expand or collapse the notifications from one app. But now we always show a notification and clicking that will cycle through all of them?
Also this will most definitely break compatibility with something like #169 but I'm sure we can work around that.

danirabbit:
> > This is how iOS handles collapsing as well and I really like that. This way you can still see the latest notification from a given app without it flooding the list. It means you can just always keep a noisy app collapsed and never really have to expand the stack if you don't want to
> 
> Sure I might need some help with CSS stuff though if we want to show like a stack with the other notifications maybe like barely visible under the topmost. Also just to be sure I understand it correctly: we keep the header with the arrow so when clicking it, it will still expand or collapse the notifications from one app. But now we always show a notification and clicking that will cycle through all of them? Also this will most definitely break compatibility with something like #169 but I'm sure we can work around that.

Thinking about it more, it would be nice just to get the current behavior in and we can think about a more complex behavior later :)

danirabbit (approved):
I'm happy with this, this looks good to me!

Marukesu (approved):
LGTM

PR #265, opened 2023-07-23
Remove `Session.vala`

Marukesu:
session storage while using the org.freedesktop.Notifications interface is flawed and broken. notifications can't be closed by apps and actions cannot be activated after the application exits or there's a system reset. so remove the between session capabilities from the indicator.

we can restore between session persistence in the server, after it has direct support to the Gtk/Portal interfaces. but for now, i think it's better to be honest here and remove it.


danirabbit:
I'm -1 on removing this functionality. There can still be valuable information in an old notification even if the notification itself can't be actioned on

Marukesu:
The idea here was more of "remove it temporary, and restore after support for the portal/GTK interfaces was added in the server". Given that the notification format is different between those interfaces and the fdo interface, a rewrite of the session file will be needed anyway.

the biggest problem with the between sessions persistence and the fdo interface are that applications cannot remove they notifications after a reboot or being closed, so the user need to be constantly cleaning-up the indicator.

teamcons:
I would then rather argue for a "Pin notification" feature to keep persistence on specific elements, noted by a pin near the "x"
or a "natural notification decay" where the notification isnt kept longer than X days, set in Privacy/Housekeeping

As of now, infinite notification persistence tend to be perceived as bug:
https://github.com/elementary/wingpanel-indicator-notifications/issues/277

It also taps into a bug which causes a main UI component to become unusable over time
https://github.com/elementary/wingpanel-indicator-notifications/issues/237
Which do not seem to have a simple solution

In its current state the user has to form the habit to manually clear notifications, else
-the whole purpose of the indicator is defeated by having relevant info lost in the sauce of irrelevant uncleared notifications
-the desktop slowly goes into unusable state
-Old useful notifications gets drown even more, so the preservation feature currently is a bit... self defeating.

PR #268, opened 2023-08-29
Make headers focusable, fix styles, add tooltip

danirabbit:
* allow focusing so we can interact via keyboard
* `image-button` class handles almost everything for us already, just make sure we remove the background when :checked
* Fix `:focus` background color
* Fix label font weight
* Add a tooltip to the arrow image

### DIFF
--- a/src/Widgets/AppEntry.vala
+++ b/src/Widgets/AppEntry.vala
@@ -52,7 +52,9 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
             name = _("Other");
         }
 
-        var image = new Gtk.Image.from_icon_name ("pan-end-symbolic", SMALL_TOOLBAR);
+        var image = new Gtk.Image.from_icon_name ("pan-end-symbolic", SMALL_TOOLBAR) {
+            tooltip_text = _("Show less")
+        };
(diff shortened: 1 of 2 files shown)


src/Widgets/AppEntry.vala
```diff
@@ -97,6 +97,12 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
         }
```
leolost2605, line 98: We probably should add something like this outside the signal after the checking for the app in `headers` (also makes the `tooltip_text` in the image construction obsolete):
`image.tooltip_text = expander.active ? _("Show less") : _("Show more");`
Also the same line could simplify the if else inside the signal.
Another thing I was wondering is whether having separated translatable strings that should be the same like this might be problematic (it's currently the same with the construction using tooltip_text = _("Show less") and the same being used in the signal)? If so we should probably put it in a method? Or maybe I'm just overthinking :)

leolost2605 (changes requested):

src/Widgets/AppEntry.vala
leolost2605: The comment is a bit unfortunately placed just below is the `expander.toggled.connect` signal handler 😅

danirabbit: I think I can even do you one better and use a custom property binding :)

leolost2605:
Very nice :) 🎉

PR #275, opened 2023-12-29
Don't create two notifications monitors

lenemter:


danirabbit:
Should notification monitor be a singleton instead?

leolost2605:
NotificationsMonitor is marked as `[SingleInstance]` which IIRC should make it a thread safe singleton automatically and returns a reference if you try to create a new one so I don't think we need this PR?
https://stackoverflow.com/questions/26832371/how-to-create-a-thread-safe-singleton-in-vala

lenemter:
Didn't know about this. Thanks @leolost2605. I'll close this pull request.

PR #283, opened 2024-10-09
Add sweep  animation

alainm23:
Just details, a sweep animation is added when clicking the clear button

https://github.com/user-attachments/assets/ae523af9-47db-4c8c-99f8-e3680d11fcfc


data/AppEntryClear.css
```diff
@@ -0,0 +1,20 @@
...
+* SPDX-FileCopyrightText: 2017-2024 elementary, Inc. (https://elementary.io)
+*/
+
+@keyframes sweep {
+    0%   { -gtk-icon-transform: rotate(0deg);   }
+    25%  { -gtk-icon-transform: rotate(30deg); }
+    50%  { -gtk-icon-transform: rotate(0deg);   }
+    75%  { -gtk-icon-transform: rotate(-30deg);  }
+    100% { -gtk-icon-transform: rotate(0deg);   }
+}
```
danirabbit, line 12: I adjusted the timing a bit and added an x transform so that the motion looks like it's pivoting from the handle more. This way you get kind of a big exaggerated sweep in the direction of the sweep as implied by the icon with a little less rebound at the end

```suggestion
@keyframes sweep {
    0%   { -gtk-icon-transform: rotate(0deg) translatex(0);   }
    20%  { -gtk-icon-transform: rotate(40deg) translatex(-5px); }
    60%  { -gtk-icon-transform: rotate(0deg)   translatex(0); }
    80%  { -gtk-icon-transform: rotate(-30deg) translatex(3px);  }
    100% { -gtk-icon-transform: rotate(0deg) translatex(0);   }
}
```

data/AppEntryClear.css
```diff
@@ -0,0 +1,20 @@
...
+
+.sweep-animation {
+    transition: all 175ms ease-in-out;
+}
+
+.sweep-animation.active {
+    animation: sweep 700ms ease-in-out;
```
danirabbit, line 19: Maybe this should be a little faster to match closer to the timeout?
```suggestion
    animation: sweep 600ms ease-in-out;
```

src/Widgets/AppEntry.vala
```diff
@@ -101,7 +110,11 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
         });
 
         clear_btn_entry.clicked.connect (() => {
-            clear (); // Causes notification list to destroy this app entry after clearing its notification entries
+            clear_btn_image.get_style_context ().add_class ("active");
+            GLib.Timeout.add (550, () => {
```
danirabbit, line 114: We should run `clear_all_notification_entries ();` before the timeout so that the response is still immediate and we're only waiting for the animation to finish to clear the header
```suggestion
            clear_all_notification_entries ();
            GLib.Timeout.add (600, () => {
```

danirabbit (changes requested):
This is such a cute idea. I love this. Added a few suggestions

alainm23:
I made the above adjustments, and also unified the stylesheets for the AppEntry widget.

src/Widgets/AppEntry.vala
```diff
@@ -71,12 +71,15 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
...
         expander_style_context.add_class ("image-button");
 
-        var clear_btn_entry = new Gtk.Button.from_icon_name ("edit-clear-all-symbolic", Gtk.IconSize.SMALL_TOOLBAR) {
+        var clear_btn_image = new Gtk.Image.from_icon_name ("edit-clear-all-symbolic", Gtk.IconSize.SMALL_TOOLBAR);
+        clear_btn_image.get_style_context ().add_class ("sweep-animation");
+
+        var clear_btn_entry = new Gtk.Button () {
             tooltip_text = _("Clear all %s notifications").printf (name)
         };
+        clear_btn_entry.add (clear_btn_image);
```
danirabbit, line 82: This should probably use the `child` property instead since `add ()` won't exist for Gtk.Button in GTK4

src/Widgets/AppEntry.vala
```diff
@@ -71,12 +71,15 @@ public class Notifications.AppEntry : Gtk.ListBoxRow {
             active = true
         };
         unowned var expander_style_context = expander.get_style_context ();
-        expander_style_context.add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
         expander_style_context.add_class ("image-button");
```
danirabbit, line 74: You need to add another style class here like `.expander` otherwise all the other toggle buttons in the panel will also get rotated icons :sweat_smile: 

Screenshot from 2024-10-09 15 30 15

danirabbit (changes requested):

src/Widgets/AppEntry.vala
alainm23: Sorry, I'm just getting familiar with architecture.

danirabbit (approved):
Love it, super cute! nice work :)

Issue #277, opened 2024-03-22
Outdated notifications lurk in the indicator


janxkoci:
### What Happened?

Very often after opening the indicator I see notifications that are days or even _weeks_ old. Notifications regularly survive reboots too. Is there any useful reason for these outdated notifications lurking around?

### Steps to Reproduce

1. Boot your device.
2. Do you already see the red dot on the notification?
3. If not, wait for some and then reboot.
4. You can see notifications surviving reboots.

### Expected Behavior

Notifications should be completely cleared by reboots at minimum. Other means to designate notifications as outdated should be considered. One such case comes to mind - if the same process in the same terminal tab sends the same notification over and over, maybe keep the last one and not each and every one of them? I have like 5 or 10 notifications telling me I closed `jupyter-lab` at the end of every work day.

It makes the whole indicator kind of useless, if I have to sift through weeks-old notifications to see what happened...

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

teamcons:
A small bandaid, to do in terminal:

empty the file where notifications are saved:
`> ~/.cache/.notifications.session ; `

then make it unwriteable, avoiding any notification saving:
`chmod -w ~/.cache/.notifications.session`


Do the latter with "+w" instead if you want to reenable notification saving across reboots.
havent tested, but the first should, as an autostart command, keep only last session.

danirabbit:
Normally it's up to apps to mark their own notifications as outdated and to make sure to replace notifications properly. 

Reboot is pretty arbitrary: a notification could be 1 second or 1 month old when rebooting.

In the specific case of Terminal, we should now replace notifications by process and clear old outdated notifications such as when a Terminal tab is focused etc.

If you have notifications that are outdated, please report against the app sending them that they need to clean up their old notifications when they become stale

janxkoci:
Oh my system with GNOME does not have this issue, so I'm fine..

PR #284, opened 2024-11-22
Allow turning off keeping notifications accross reboots

teamcons:
this allows disabling the Session feature for those who do not want it
There is no toggle in the UI - It isnt a proposed change to desktop experience, only a tweak.

this could be an answer for those who perceive notification persistence as a bug/undesirable: https://github.com/elementary/wingpanel-indicator-notifications/issues/277

or ignore that part of the desktop until it becomes a problem:
https://github.com/elementary/wingpanel-indicator-notifications/issues/237

only new notifications do not get saved, when you toggle this to false. The already saved ones are kept.


teamcons:
Pfft i thought it worked but it doesnt

PR #299, opened 2025-10-04
Introduce keep-notifications key

teamcons:
introduce a key to check at saving time whether to save or not, allowing to control whether to keep notification accross reboots

the DE should control and limit behaviours of apps who do not tidy up behind them.
the setting could be exposed in the UI in the privacy settings

Not checking at load time because it would mean notifications are still stored despite user not wanting to.


teamcons:
> I'm personally -1 on this. I really don't like doing workarounds for misbehaving apps. It prevents the problem from actually being solved at the source and just kicks the can downstream instead

It is unrealistic to assume all apps behave, tidy up behind themselves, and dont leave informations behind them
Theres reasons apps cant place themselves, cant keep_on_top themselves, and cant record the whole screen at all time themselves.

> But more importantly session end is a really arbitrary time frame. If a housekeeping feature for notifications should exist, it should behave like other housekeeping settings and clean up notifications some specific duration after they were sent
> 
> If you just want to arbitrarily clear all notifications there is already a clear notifications button

I think a handy DE is one that doesnt require me to manually tidy it regularly
(not even mentioning the years old bug which has wingpanel going slower over time if notifications arent emptied)

session end is when user is finished with the computer.

> 
> If you just want to arbitrarily clear all notifications there is already a clear notifications button

Why then does these exist in privacy if we can do it manually?

image

teamcons:
> > It seems like this branch is trying to workaround that some apps aren't withdrawing their motivations when they are no longer relevant and/or the bug that prevents notifications from a previous session from being withdrawn.

it is also working around the fact that the DE is piling up stuff over time
(and borking itself as a byproduct too)


> @teamcons To me this doesn't feel like a good workaround because session restart is arbitrary. It isn't related to whether or not the notification is still relevant and unlike other housekeeping settings, it isn't related to an actual amount of time that has passed since the notification was sent. Session restart could happen when a notification is 30 days old or 30 seconds old.

I repeat my statement about session start or end:
> session end is when user is finished with the computer.

which makes it relevant for the DE to clear session stuff that is irrelevant because the user has decided to terminate the session.
thats like saying doing package updates on shutdown is arbitrary

it is a better workaround than nothing at all. Several people have come forward to express them not wanting this behaviour, over time.

I am offering a patchup that does the minimum amount of change to keep people happy, to let the default unchanged, and have a DE that doesnt bork itself.

I _can_ come up with a more complex solution to add cleanup in the settings-daemon, and parse the session file in cache, but it would have been much better to tell me "id rather have that" from the start - arguing eats away dev time that is offered to you for free, and leaves a bad taste that doesnt make me wanna offer it again.

PR #306, opened 2026-08-26
NotificationsList: use GLib.ListStore

danirabbit:
Closes #169 

Next step towards ListView: use a ListStore and bind model. Create headers in the header func

### DIFF
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -99,16 +99,16 @@ public class Notifications.NotificationsList : Granite.Bin {
             return;
         }
 
-        var app_entry = app_entries[row_entry.notification.desktop_id];
+        var app_entry = app_entries[row_app_id];
         if (app_entry != null) {
             app_entry.add_notification_entry (row_entry);
 
         } else {
-            var app_entry = new AppEntry (row_entry.notification.app_info);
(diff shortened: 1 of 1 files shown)


src/Widgets/NotificationsList.vala
```diff
@@ -95,51 +99,37 @@ public class Notifications.NotificationsList : Granite.Bin {
...
-
-    public async void add_entry (Notification notification) {
-        var entry = new NotificationEntry (notification);
-        listbox.append (entry);
-
-        if (app_entries[notification.desktop_id] != null) {
-            var app_entry = app_entries[notification.desktop_id];
-            app_entry.add_notification_entry (entry);
+        if (app_entries[row_entry.notification.desktop_id] != null) {
+            var app_entry = app_entries[row_entry.notification.desktop_id];
```
lenemter: You should avoid doing a hashtable lookup twice:


```suggestion
        var app_entry = app_entries[row_entry.notification.desktop_id];
        if (app_entry != null) {
```

lenemter: Also `row_entry.notification.desktop_id` seems to already be cached in `row_app_id`

lenemter (changes requested):

lenemter (approved):

PR #307, opened 2026-08-26
AppEntry: don't hold references to Notifications

danirabbit:
This makes sure that AppEntry is as dumb as possible and not referencing NotificationEntries all over the place


src/Widgets/NotificationsList.vala
```diff
@@ -25,7 +25,7 @@ public class Notifications.NotificationsList : Granite.Bin {
     private GLib.HashTable<string, GLib.DateTime> app_datetime;
     public Gee.HashMap<string, AppEntry> app_entries { get; private set; }
 
-    private ListStore list_store;
+    public ListStore list_store { get; private set; }
```
lenemter: You shouldn't expose ListStore like that since it's modifiable (has append/remove methods). Instead you should keep ListStore private and expose it as a ListModel:

```
private ListStore list_store;
public ListModel notification_entries { get { return list_store; } }
```

src/Widgets/NotificationsList.vala
```diff
@@ -107,13 +107,31 @@ public class Notifications.NotificationsList : Granite.Bin {
...
+                var entry = (NotificationEntry) list_store.get_item (i);
+                if (entry.notification.desktop_id == row_app_id) {
+                    entries++;
+                }
+            }
+
+            if (entries == 0) {
+                clear_app_entry (app_entry);
+            }
+        });
```
lenemter: Why is this here and not in `add_entry`?

danirabbit:
Gonna draft this and break it up

src/Widgets/NotificationsList.vala
```diff
@@ -169,6 +168,23 @@ public class Notifications.NotificationsList : Granite.Bin {
...
+        for (int i = 0; i < list_store.n_items; i++) {
+            var notification_entry = (NotificationEntry) list_store.get_item (i);
+            if (notification_entry.notification.desktop_id == entry.notification.desktop_id) {
+                entries++;
+            }
+        }
+
+        if (entries == 0 && app_entries[entry.notification.desktop_id] != null) {
+            clear_app_entry (app_entries[entry.notification.desktop_id]);
+        }
```
danirabbit: I wonder if we could instead check if the app_entry has a parent?

danirabbit:
@lenemter can I get a re-review please :)

lenemter (approved):

PR #308, opened 2026-08-26
NotificationEntry: bind hide to settings

danirabbit:
Make sure we're using settings as the source of truth and not UI. This is admittedly more complicated seeming but when we are using ListView it will make it easier to recycle list items

### DIFF
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -244,7 +244,7 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
 
         settings.bind_with_mapping (
             "headers", revealer, "reveal-child", GET,
-            (SettingsBindGetMappingShared) get_bind_func, () => false,
+            get_bind_func, () => false,
             new Variant.string (notification.desktop_id), null
         );
 
--- a/src/Widgets/NotificationEntry.vala
+++ b/src/Widgets/NotificationEntry.vala
@@ -14,9 +14,9 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
     private const int ICON_SIZE_PRIMARY = 48;
     private const int ICON_SIZE_SECONDARY = 24;
 
-    private static Settings settings;
     private static Regex entity_regex;
     private static Regex tag_regex;
+    private static Settings settings;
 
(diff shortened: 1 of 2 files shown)


src/Widgets/NotificationEntry.vala
```diff
@@ -14,6 +14,7 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
     private const int ICON_SIZE_PRIMARY = 48;
     private const int ICON_SIZE_SECONDARY = 24;
 
+    private static Settings settings;
     private static Regex entity_regex;
     private static Regex tag_regex;
```
ryonakano: I think `setting` should be declared after `tag_regex` to match the order in `static construct`.

src/Widgets/NotificationEntry.vala
```diff
@@ -239,6 +242,18 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
             delete_revealer.reveal_child = false;
         });
 
+        settings.bind_with_mapping (
+            "headers", revealer, "reveal-child", GET,
+            (value, variant, user_data) => {
+                var app_id = ((Variant) user_data).get_string ();
+                var headers_table = (HashTable<string, bool>) variant;
+                value.set_boolean (headers_table[app_id ]);
```
ryonakano:
```suggestion
                value.set_boolean (headers_table[app_id]);
```

Extra whitespace.

src/Widgets/NotificationEntry.vala
```diff
@@ -239,6 +242,18 @@ public class Notifications.NotificationEntry : Gtk.ListBoxRow {
...
         });
 
+        settings.bind_with_mapping (
+            "headers", revealer, "reveal-child", GET,
+            (value, variant, user_data) => {
+                var app_id = ((Variant) user_data).get_string ();
+                var headers_table = (HashTable<string, bool>) variant;
+                value.set_boolean (headers_table[app_id]);
+                return true;
+            },
```
lenemter: Can we extract this to a static method please? These kinds of lambdas are prone to cause memory leaks

ryonakano (approved):
Works confirmed using `notify-send`. LGTM if Leo's comment is addressed.

PR #309, opened 2026-08-27
NotificationEntry: construct once then set gicon

danirabbit:
When we switch to ListView we're going to want `notification` to be a settable property so that we can recycle widgets. This makes it so we construct the Images once and then set gicons.

Fixes an issue where we sometimes got a checkerboard with image missing instead of just showing the app icon and where notify-send notifications had image missing


jeremypw (approved):
Makes sense as part of a roadmap.

PR #312, opened 2026-08-27
Indicator: show accels if available

danirabbit:
Goes with https://github.com/elementary/wingpanel/pull/699


jeremypw (approved):
Code looks good and works as expected.

I wonder whether the tooltip would look more balanced if the accels were right-aligned inline with the description rather the usual position for a twoline tool tip?  Not blocking though.

PR #315, opened 2026-08-28
AppEntry: use latest Granite features

danirabbit:
* Granite.HeaderLabel
* Granite.Box


src/Widgets/AppEntry.vala
```diff
@@ -3,7 +3,7 @@
 * SPDX-FileCopyrightText: 2015-2026 elementary, Inc. (https://elementary.io)
 */
 
-public class Notifications.AppEntry : Gtk.ListBoxRow {
+public class Notifications.AppEntry : Granite.Bin {
```
zeebok, line 6: Changing this type won't affect the performance improvements you've been working for with the ListView, right?

zeebok (approved):

src/Widgets/AppEntry.vala
danirabbit: Yeah we're using ListView now and not ListBox, so we actually don't want to use ListBox types anymore

PR #321, opened 2026-09-01
NotificationEntry: remove from list

### LINKED ISSUE #317: Notification indication and tooltip is not updated when clearning notifications
### What Happened?

Regression of f8a676e65cf51453a43f50aed0e4568632fc02f2.

The notification indication in the panel icon and the number of notifications in the tooltip is not updated when clearning notifications.

### Steps to Reproduce

1. Build and install from the source code at f8a676e65cf51453a43f50aed0e4568632fc02f2
2. Run `killall io.elementary.wingpanel && sleep 5 && notify-send foobar`
3. Click the close button in the notification bubble or the clear button in the popover
4. See the notification indication on the bell mark icon and the number of notifications in the tooltip is not updated  
  Image

5. Build and install from the source code at 43fbd50e2a1e0f6d3d42e492708e98aeed193706 (the parent commit of the above commit)
6. Reproduce the step 2 and 3
7. See the notification indication on the bell mark icon and the number of notifications in the tooltip is updated

### Expected Behavior

The notification indication on the bell mark icon and the number of notifications in the tooltip should be updated as before.

### OS Version

Early Access

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Early Access or Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

danirabbit:
Fixes #317
Make sure that the actual notification data is removed, not just the row from the listbox

### DIFF
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -107,7 +107,7 @@ public class Notifications.NotificationsList : Granite.Bin {
 
     public async void add_entry (Notification notification) {
         var entry = new NotificationEntry (notification);
-        entry.remove.connect ((_entry) => remove_notification (_entry.notification));
+        entry.remove.connect (remove_notification);
 
         list_store.insert_sorted (entry, sort_func);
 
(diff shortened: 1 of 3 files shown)


src/Widgets/NotificationsList.vala
```diff
@@ -157,6 +159,16 @@ public class Notifications.NotificationsList : Granite.Bin {
         }
     }
 
+    private void remove_notification (Notification notification) {
```
jeremypw: Is there a reason we cannot pass the entry itself as a parameter and then use `list_store.find (entry, out pos)` to get its position rather than explicitly looping through the items?  It seems to work for me although I didn't check for memory leaks.

danirabbit: I think I was just working late and tired lmao

jeremypw (approved):

PR #322, opened 2026-09-02
NotificationsList: add get_n_app_items

danirabbit:
There's a couple cases where we don't actually need to count apps, we just want to know if there's any notifications, so fix that

I both:
* Want to use the ListStore as the source of truth
* Want to eventually not hold references to list headers, so this is a step towards that

### DIFF
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -158,7 +158,7 @@ public class Notifications.NotificationsList : Granite.Bin {
 
         Session.get_instance ().remove_notifications (to_remove);
 
-        if (app_entries.size == 0) {
+        if (get_n_app_items () == 0) {
             Session.get_instance ().clear ();
         }
     }
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -132,12 +132,8 @@ public class Notifications.NotificationsList : Granite.Bin {
 
     public uint get_n_app_items () {
         var app_list = new GenericSet<string> (str_hash, str_equal);
-
         for (var i = 0; i < list_store.n_items; i++) {
-            var app_id = ((NotificationEntry) list_store.get_item (i)).notification.desktop_id;
-            if (!app_list.contains (app_id)) {
-                app_list.add (app_id);
-            }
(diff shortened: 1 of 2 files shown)


src/Widgets/NotificationsList.vala
```diff
@@ -130,6 +130,19 @@ public class Notifications.NotificationsList : Granite.Bin {
...
+
+        for (var i = 0; i < list_store.n_items; i++) {
+            var app_id = ((NotificationEntry) list_store.get_item (i)).notification.desktop_id;
+            if (app_list.find (app_id) == -1) {
+                app_list.append (app_id);
+            }
+        }
+
+        return app_list.n_items;
+    }
```
lenemter, line 140: Gtk.StringList is unnessesary complicated for this usecase. You should use something like GLib.GenericSet as it's more effective and much simpler

src/Widgets/NotificationsList.vala
```diff
@@ -131,16 +131,16 @@ public class Notifications.NotificationsList : Granite.Bin {
...
 
     public uint get_n_app_items () {
-        var app_list = new Gtk.StringList ({});
+        var app_list = new GenericSet<string> (str_hash, str_equal);
 
         for (var i = 0; i < list_store.n_items; i++) {
             var app_id = ((NotificationEntry) list_store.get_item (i)).notification.desktop_id;
-            if (app_list.find (app_id) == -1) {
-                app_list.append (app_id);
+            if (!app_list.contains (app_id)) {
```
jeremypw: Doesn't a set automatically ensure unique entries? Is this test necessary?

danirabbit: Oh I didn't know that! So cool! This is my new favorite collection type :heart_eyes:

jeremypw: Yeah, just need to be careful when removing notifications - can't just remove the corresponding app_id I don't think. Need to check if the collection keeps track of how many times the id has been added/removed. If not, have to rebuild the set or otherwise decide if the app_id can be removed.

jeremypw: Looks like the number of apps is counted each time it is needed so shouldn't be a problem at the moment.

jeremypw (approved):

PR #323, opened 2026-09-02
AppEntry: remove app_notifications

danirabbit:
Remove a round trip where `NotificationEntry.clear ()` → `AppEntry.remove_notification_entry ()` → `NotificationEntry.dismiss ()` can just be NotificationEntry calling dismiss internally

Moving counting and unsetting headers from `NotificationEntry.clear ()` → `AppEntry.remove_notification_entry ()`  to `NotificationEntry.remove ()` → `NotificationsList.remove_notification ()`
*  This uses a FilterListModel to count the items for a given `app_id` from the `ListStore` which is our source of truth
* We also get to remove unsetting headers from `clear_app_entry` because we already call `NotificationEntry.dismiss ()` → `NotificationEntry.remove ()` → `NotificationsList.remove_notification ()`

Now that we count up `items_for_appid` when removing, call `clear_app_entry ()` here using the handy `HashMap` of `app_id`s.

Remove the list of notifications from AppEntry.

Finally, remove the unused `clear ()` signal and make sure external calls are to the function `dimiss ()` instead

Rebase merge if desired

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -179,7 +179,7 @@ public class Notifications.Indicator : Wingpanel.Indicator {
             var entry = (NotificationEntry) nlist.notification_items.get_item (i);
             if (id == entry.notification.server_id) {
                 entry.notification.server_id = 0; // Notification is now outdated
-                entry.clear ();
+                entry.dismiss ();
                 return;
             }
         }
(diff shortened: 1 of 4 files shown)


zeebok (approved):
What a web of calls! This seems good to me, definitely cleaner 😅

PR #324, opened 2026-09-02
NotificationEntry: connect to server_id

danirabbit:
Instead of directly dismissing the notification row widget, we should mark these notifications as outdated and then let the row listen to that property. This is going to come in handy to make our `NotificationsList` `ListStore` be a list of `Notifications` and not `NotificationEntry`s, a pre-req for performant `ListViews`


jeremypw (approved):
Makes sense and works as expected.

PR #326, opened 2026-09-02
NotificationEntry: start/stop timestamp timer on map/unmap

danirabbit:
This timer doesn't need to be running while the notification is unmapped


jeremypw (approved):
Makes sense and works as expected.

PR #330, opened 2026-09-03
NotificationEntry: create bind function

danirabbit:
ListView will want the widget construction to be separated from setting the notification. Move everything to a bind function so widget construction happens separately


jeremypw (approved):

PR #338, opened 2026-09-03
NotificationList: find notification when removing, not entry

danirabbit:
Fixes a bug that was introduced when we switch the listmodel from entries to notifications


src/Widgets/NotificationsList.vala
```diff
@@ -158,12 +158,13 @@ public class Notifications.NotificationsList : Granite.Bin {
     }
 
     private void remove_notification (NotificationEntry notification_entry) {
```
zeebok: Since NotifictionEntry is now not used at all in this method, should we just change the signature to take in a Notification?

zeebok (approved):

PR #343, opened 2026-09-04
NotificationsList: consume all view widgets

danirabbit:
Another end game one, make sure indicator is no view, only model and controller

### DIFF
--- a/src/Widgets/NotificationsList.vala
+++ b/src/Widgets/NotificationsList.vala
@@ -31,7 +31,7 @@ public class Notifications.NotificationsList : Granite.Bin {
         var not_disturb_switch = new Granite.SwitchModelButton (_("Do Not Disturb"));
 
         var dnd_switch_separator = new Gtk.Separator (HORIZONTAL) {
-            margin_top = 3,
+            margin_top = 3
         };
 
         var clear_all_btn_separator = new Gtk.Separator (HORIZONTAL) {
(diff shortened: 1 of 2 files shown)


src/Widgets/NotificationsList.vala
```diff
@@ -71,7 +90,17 @@ public class Notifications.NotificationsList : Granite.Bin {
         stack.add_named (placeholder, "placeholder");
         stack.add_named (scrolled, "list");
 
-        child = stack;
+        var main_box = new Gtk.Box (VERTICAL, 0) {
+            width_request = 360
+        };
+        main_box.append (not_disturb_switch);
+        main_box.append (dnd_switch_separator);
+        main_box.append (scrolled);
```
zeebok: Should this be stack instead?

src/Widgets/NotificationsList.vala
```diff
@@ -27,6 +28,24 @@ public class Notifications.NotificationsList : Granite.Bin {
     }
 
     construct {
+        var not_disturb_switch = new Granite.SwitchModelButton (_("Do Not Disturb"));
```
zeebok, line 31: This was styled in Indicator so not sure if it needs some styling here?

src/Widgets/NotificationsList.vala
danirabbit: Oops yes. Good catch!

src/Widgets/NotificationsList.vala
danirabbit: Yeah I'm not sure why it was before? This is not a list header, it's just a toggle so it doesn't make sense to style it as a header imo

zeebok (approved):

PR #354, opened 2026-09-04
Notification: scope and sort properties

danirabbit:


zeebok (approved):
Oh wow I wasn't even thinking about property order when looking through all of those PRs 😅

PR #359, opened 2026-09-07
Indicator: add clear-app action

danirabbit:


vjr (approved):
I'm unable to review (unfamiliar with the "action" aspects of) the code, **_seems_** OK to me, and fixes the issue with my quick tests, I think this can be merged.

The issue here https://github.com/elementary/panel-notifications/issues/328#issuecomment-5564157851 seems to still happen IFF the collapse/expand triangle was already expanded state before I sent test notifications and popped up the indicator but this might be a separate issue / already addressed by one of the other PRs.

jeremypw (approved):
Code looks good and works as expected.

PR #360, opened 2026-09-11
Indicator: create header settings if it doesn't exist

danirabbit:
When we get a new notification, make sure that we create visibility settings before trying to bind to it in items and headers

### DIFF
--- a/src/Indicator.vala
+++ b/src/Indicator.vala
@@ -154,6 +154,14 @@ public class Notifications.Indicator : Wingpanel.Indicator {
             sort_list_model.section_sorter.changed (DIFFERENT);
         }
 
+        // Create value if it doesn't exist and set true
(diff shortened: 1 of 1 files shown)


src/Indicator.vala
```diff
@@ -154,6 +154,14 @@ public class Notifications.Indicator : Wingpanel.Indicator {
             sort_list_model.section_sorter.changed (DIFFERENT);
         }
 
+        // Create value if it doesn't exist and set true
+        var settings = new Settings ("io.elementary.panel.notifications");
+        var headers = (HashTable<string, bool>) settings.get_value ("headers");
+        if (!headers.contains (notification.desktop_id)) {
```
flodavid, line 160: `headers` can be null here I think, so we should check for nullability ?

